PR #7, opened 2026-02-02
Add deep interactive exploration mode for researcher

IovetsNikolay:
performInteractiveExploration() purpose: Discover what each button/link actually DOES when clicked. 
This could be baseline  logic, after proof that it researches better

Algorithm:
1. Collect all interactive elements from ARIA snapshot
   - No AI call needed - direct parsing
2. For each element:
   - Click it
   - Detect result (navigation, modal, content change)
   - Record outcome
   - Return to original state
3. Generate structured report
 
Flow:
/research --deep

1. performDeepAnalysis() runs first:
   └── Expands hidden menus, dropdowns, accordions
   └── Documents what's inside them
2. performInteractiveExploration() runs after:
   └── Clicks visible buttons/links
   └── Documents what each one does
3. Combined output:
   └── Complete picture of page structure AND behavior

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,9 @@ out/
 # Example folder
 example/
 
+# Knowledge (may contain credentials)
+knowledge/
+
 # Output directories
 output/
 output/logs/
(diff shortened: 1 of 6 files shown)


DavertMik:
Implemented here https://github.com/testomatio/explorbot/commit/60dd649b855161fe40af0985fb91fa0506b657aa#diff-2aeca8cdfa6a17b18839b93b928d9837fcd55364237a8c1fe9d6c297b57154db

PR #17, opened 2026-04-05
Update docs according to CHANGELOG.md

mi4i0:


DavertMik:
Outdated. Update and resend if needed

PR #41, opened 2026-05-27
Add interactive mode

DenysKuchma:
## Summary                                                                                                                                        
This PR adds an interactive mode to `doc-collector` so page documentation can include verified UI behavior, not just static page structure.
The new mode explores selected interactions, captures state changes, and falls back safely to static documentation when interactive results are weak or unreliable.

 ## What changed                                                                                                                                                                                                                                                                                                                                                      
  - added opt-in `interactive` mode for `doc-collector` 
  - added deterministic interaction exploration for:
   - content/detail links 
   - pagination
   - tab-like controls
   - meaningful buttons 
  - added `State Transitions` output in page docs using a `before -> action -> after` format
  - added support for carrying interaction-discovered URLs into the crawl queue
  - added output normalization and coverage notes for weak or partial documentation
  - improved fallback behavior when interactive or schema generation fails 
  - cleaned up the interactive implementation and removed legacy / dead code paths

### DIFF
--- a/boat/doc-collector/src/ai/documentarian.ts
+++ b/boat/doc-collector/src/ai/documentarian.ts
@@ -1,19 +1,35 @@
 import dedent from 'dedent';
 import { z } from 'zod';
 import type { AIProvider } from '../../../../src/ai/provider.ts';
+import type Explorer from '../../../../src/explorer.ts';
 import type { WebPageState } from '../../../../src/state-manager.ts';
+import { tag } from '../../../../src/utils/logger.ts';
 import type { DocbotConfig } from '../config.ts';
+import { collectDocInteractions } from './tools.ts';
 
 class Documentarian {
   private provider: AIProvider;
   private config: DocbotConfig;
+  private explorer?: Explorer;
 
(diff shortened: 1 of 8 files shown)


boat/doc-collector/src/ai/documentarian.ts
```diff
@@ -151,19 +382,66 @@ class Documentarian {
...
+
+  private hasMenuSection(research: string): boolean {
+    return /##\s+(menu|navigation|header)/i.test(research);
+  }
+
+  private hasPrimaryContentEvidence(research: string): boolean {
+    return /##\s+(content|cards|results|grid|catalog)/i.test(research);
+  }
+
+  private looksLikeItemAction(action: string, url: string): boolean {
```
DavertMik: too many hardcoded solutions

DavertMik (changes requested):
Same for new features too.
See CLAUDE.md

Image

Regexes are fine for some quick wins but evaluating actions by categories is what LLM must do (it does it well). If a solution involves Regex - probably it is not a solution. It might be optimization but not a solution!

Issue #33, opened 2026-05-14
Logs improvements


DavertMik:
- We need to improve UI/UX of logs in TUI mode
- what information is logged
- how it is structured
- what users want to see
- what information is less needed

review different testing flows and propose improvements

- for TUI
- for non interactive mode

Polish final output it is overslopped:

<img width="1601" height="911" alt="Image" src="https://github.com/user-attachments/assets/bbce7826-64b5-48a9-9bd8-83ebd7f67e8c" />

```
┌────┬────────┬───────────────────────────────────────────────────────────────────────────────┬───────────┬────────┬───────┐
│ #  │ Status │ Title                                                                         │ Priority  │ Time   │ Steps │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 1  │ failed │ Create a new requirement using the File tab and verify it appears in the list │ critical  │ 227.8s │ 48    │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 2  │ failed │ Edit an existing requirement and verify the changes persist                   │ critical  │ 143.9s │ 37    │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 3  │ passed │ Create a requirement and then delete it to verify delete functionality        │ important │ 144.1s │ 34    │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 4  │ passed │ Switch to Linked tab and verify requirements filter correctly                 │ normal    │ 49.8s  │ 5     │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 5  │ passed │ Switch to Orphaned tab and verify filtering works                             │ normal    │ 36.3s  │ 5     │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 6  │ passed │ Toggle display mode to view requirements by suites                            │ normal    │ 32.2s  │ 9     │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
│ 7  │ passed │ Switch to Jira tab and verify Jira integration features                       │ normal    │ 30.3s  │ 5     │
├────┼────────┼───────────────────────────────────────────────────────────────────────────────┼───────────┼────────┼───────┤
[...]

PR #51, opened 2026-06-04
 Improve TUI and CLI log output

DenysKuchma:
## Changes                                                                                                                                                                                        
https://github.com/testomatio/explorbot/issues/33                                                                                                                                                                           
  - Replaces the large final results table with a compact summary.                                                                                                                                  
  - Shows failed and slow tests only when useful.                                                                                                                                                   
  - Groups screencast output by folder/pattern instead of listing every file.                                                                                                                       
  - Cleans noisy Testomat.io reporter console output.                                                                                                                                               
  - Hides low-value internal substeps from console/TUI logs.                                                                                                                                        
  - Limits long multiline blocks in TUI with context-aware truncation.                                                                                                                              
  - Shortens live Pilot verdict output while preserving full summaries in reports.                                                                                                                  
  - Fixes final result aggregation when execution continues across planning styles.


src/utils/log-filters.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (entry.type !== 'substep') return false;
+  return isLowValueSubstep(entry.content);
+}
+
+export function isLowValueTuiLog(entry: LogFilterEntry): boolean {
+  if (entry.type !== 'substep') return false;
+  return isLowValueSubstep(entry.content);
+}
+
+function isLowValueSubstep(content: string): boolean {
```
DavertMik: i think we can add another log level for this instead of pushing everything to substep
like `operation` and list all low level operations

src/utils/log-filters.ts
```diff
@@ -0,0 +1,50 @@
+interface LogFilterEntry {
+  type: string;
+  content: string;
+}
+
+export class RecentStepFilter {
```
DavertMik, line 1: better to use enother log type instead

src/ai/pilot.ts
```diff
@@ -1020,3 +1020,11 @@ export class Pilot implements Agent {
     `;
   }
 }
+
+function summarizePilotReason(reason: string): string {
```
DavertMik: we must pass these contraints to pilot itself
cutting string wont be sufficient

src/components/LogPane.tsx
```diff
@@ -17,12 +17,20 @@ interface LogPaneProps {
   verboseMode: boolean;
 }
 
+function getMultilineLimit(content: string): number {
+  if (content.startsWith('Results:')) return 24;
+  if (content.startsWith('Plan:') || content.startsWith('Generated tests') || content.startsWith('Screencasts')) return 18;
+  if (content.startsWith('# Session Analysis')) return 22;
+  if (content.startsWith('ARIA Snapshot:') || content.startsWith('HTML Content:')) return 10;
+  if (content.includes('| type |') || content.includes('| Type |')) return 12;
```
DavertMik: no hardcoding pleeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeease

DavertMik: maybe we can add option to multiline log ?

DavertMik (changes requested):
- add another logtype
- add log type params (like length?)
- add prompt restriction pilot

avoid regex, startsWith, string catching solutions 

they are hard to detect, understand, add, and refactor
we must easily add new logs in new features and not to hardcode them in future!

src/utils/reporter-console.ts
```diff
@@ -0,0 +1,46 @@
...
+    return await fn();
+  } finally {
+    console.log = originalLog;
+    console.warn = originalWarn;
+  }
+}
+
+export function cleanReporterLine(args: any[]): string | null {
+  const text = stripAnsiText(args.map((arg) => String(arg)).join(' '));
+  if (text.includes('Testomatio Reporter v')) return null;
```
DavertMik: ?????

we have output formatting options in reporter!

see `--format` optoin in reporter or env vars 
it allows supressing output
actually we can do whatever we need in reporter

DO NOT HARDCODE TEXTS LIKE THIS!

DavertMik (changes requested):
Do not hardcode reporter texts
Use reporter output format or other option from reporter to ensure we hide texts 

you can use https://github.com/testomatio/reporter/blob/2.x/docs/configuration.md#testomatio_log_level

or --format option

DavertMik:
image

1. what is max-tests in footer?

DavertMik:
image
blue footer is not nice, we should remove it or think how we can show it in a better way

DavertMik:
image
when test started you must print info about it
as it is unclear that test has fnished and what have started next?
add extra space

```

Statring test XXXXX
```

DavertMik (approved):

PR #57, opened 2026-06-17
Feat/improve explorbot planning and verification

DenysKuchma:
## Summary
                                                                                                                                                                                                 
  Improves Explorbot planning, verification, and tool usage during exploratory runs.                                                                                                             
                                                                                                                                                                                                 
  ### Changes                                                                                                                                                                                     
                                                                                                                                                                                                 
  - Added session analysis after `explorbot test`.                                                                                                                                               
  - Improved exact quoted-value verification.                                                                                                                                                    
  - Added `hover` tool for hover-only UI controls.                                                                                                                                               
  - Improved planning around stable list pages, filters, tabs, and visible-only values.                                                                                                          
  - Improved handling of explicit `do not create/edit/delete` restrictions.                                                                                                                      
  - Clarified tool-name and tool-schema usage for agents.                                                                                                                                        
  - Reduced false-positive finishes with negative assertions.

### DIFF
--- a/src/ai/fisherman.ts
+++ b/src/ai/fisherman.ts
@@ -9,6 +9,7 @@ import { loop } from '../utils/loop.ts';
 import type { Agent } from './agent.ts';
 import { type FishermanResult, createFishermanTools } from './fisherman-tools.ts';
 import type { Provider } from './provider.ts';
+import { dataProtectionRules } from './rules.ts';
 
 const MAX_ITERATIONS = 15;
 const MAX_TOOL_ROUNDTRIPS = 5;
@@ -85,7 +86,7 @@ export class Fisherman implements Agent {
       baseEndpoint: this.baseEndpoint,
     });
 
-    const conversation = this.provider.startConversation(this.buildSystemPrompt(endpointList, scopeUrl), 'fisherman');
+    const conversation = this.provider.startConversation(this.buildSystemPrompt(endpointList, Object.keys(tools), scopeUrl), 'fisherman');
     conversation.addUserText(this.buildTaskPrompt(instructions));
 
     await loop(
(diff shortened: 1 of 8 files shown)


src/ai/fisherman.ts
```diff
@@ -197,13 +197,21 @@ export class Fisherman implements Agent {
       ${endpointList}
       ${scopeBlock}
 
+      AVAILABLE TOOLS:
+      getEndpointSpec, request, finish, stop.
```
DavertMik: can we have tool list generated to ensure we won't foget to update them here

src/ai/fisherman.ts
```diff
@@ -217,6 +225,9 @@ export class Fisherman implements Agent {
 
       ${instructions}
 
+      If these instructions conflict with an explicit no-mutation restriction, do not create data.
```
DavertMik: let's add these information into rules.ts like this 

https://github.com/testomatio/explorbot/blob/main/src/ai/rules.ts#L141

So let's combine cases:

- prohibit of deletion existing cases (only through created ones, or using fisherman)
- prohibt mutating
- prohibit using fisherman

I would combine all those rules there and inject here as `dataProtectionRules`

And have one place where we define our rule policy
And allow user to change it when needed

src/ai/planner.ts
```diff
@@ -341,6 +344,12 @@ export class Planner extends PlannerBase implements Agent {
...
+      If the list is empty or no concrete item names are visible, do not invent "known" or "existing" items. Prefer empty-state, no-match search, clear-search, or read-only list behavior scenarios.
+      Do not use transient create/edit/new modal or form URLs as scenario startUrl unless the scenario specifically tests that form. For unrelated scenarios, start from the stable parent list/detail page and open the form during the test if needed.
+      Search, filter, sorting, tab, and list inspection scenarios must start from a stable list/index page where those controls are visible, not from a create/edit form.
+      When the user focus is search, filter, sorting, tabs, or list inspection, do not add create/update/delete setup or cleanup to the same scenario unless the user explicitly requested that data-changing workflow. Use existing visible data or valid empty-state checks instead.
+      For dropdowns, filters, tabs, run types, statuses, and other option sets, only propose values explicitly visible in research, visited pages, or prior observed flows. Do not infer common options that are not present.
```
DavertMik: Can be shortened to 1-2 lines

DavertMik (changes requested):
Minor adjustments

DenysKuchma:
@DavertMik Also add:
1) Prevents Explorbot from planning or passing entity-specific actions when the action is not confirmed for that exact UI entity type
2) Reduces long no-progress runs by failing earlier when repeated attempts or execution errors do not change the browser state

PR #59, opened 2026-06-22
Handle transient navigation errors before browser recovery

DenysKuchma:
## Summary                                                                                                                                                                                 

  Prevents transient SPA navigation errors from immediately triggering browser recovery and page reloads.                                                                                    
                                                                                                                                                                                             
  When Playwright throws navigation-related context errors during page capture or state verification, Explorbot now waits for the page DOM/URL to settle and retries before falling back to  
  browser recovery.                                                                                                                                                                          

  - Added shared browser error classification for navigation transition errors
  - Retried page state capture when the page is still navigating
  - Added a wait-and-retry step before browser recovery for transient navigation failures
  - Added unit coverage for browser error classification


DavertMik (approved):

DavertMik (changes requested):
I smell overengineering here
Playwright must already handle situations like loading DOM tree with doc content loaded, so I don't think we need to re-implement it

HOWEVER, if you open a project in testomat.io there is a page spinner and content is not loaded, why dom is loaded

so the fix you need must be smarter:

to understand by XHR network requests and spinner classes on page if it is in loading mode. 

please USE playwright api, do not re-implement it!

wait for all XHR requests settled on page load || all spinner classes to be hidden (spinner classes can be configured) || timeout

```js
Promise.race(
   xhr requests.
   timeout,
   spinners
)
```

DavertMik (approved):

PR #60, opened 2026-06-28
Reuse successful test experience during execution

DenysKuchma:
## Summary                                                                                                                                                                                    
  - wire experience retrieval into Tester and Pilot execution context                                                                                                                           
  - preserve full tool execution outputs through conversation compaction                                                                                                                        
  - unwrap AI SDK JSON tool results before historian processing                                                                                                                                 
  - save reusable ACTION experience from successful fallback attempts                                                                                                                           
  - prioritize saved experience commands before new fallback locators

### DIFF
--- a/src/ai/tester.ts
+++ b/src/ai/tester.ts
@@ -581,6 +581,7 @@ export class Tester extends TaskAgent implements Agent {
     }
 
     if (isNewUrl) {
+      const experience = this.getExperience(currentState);
       const alreadySeenUiMap = this.seenUiMapUrls.has(currentUrl);
       let research = '';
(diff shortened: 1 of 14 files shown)


src/ai/tester.ts
```diff
@@ -883,6 +888,8 @@ export class Tester extends TaskAgent implements Agent {
       ${this.buildAvailableFiles()}
 
       ${knowledge}
+
+      ${experience}
```
DavertMik, line 892: no, this can quickly overflow all context!

DavertMik (changes requested):
pushing all experience is not an option as we might get irrelevant or non working experience there

PR #61, opened 2026-06-30
Improve form state verification and split-pane UI awareness

DenysKuchma:
## Summary 

  - Serialize live form control state into HTML snapshots so filled inputs can be verified from page diffs.
  - Mark elements as covered/offscreen/target using browser hit-testing metadata.
  - Teach researcher output to treat split-pane detail panels as active UI context instead of mixing them with list/toolbar sections.                                        
  - Improve inline create/edit guidance for icon-only confirm controls near edited fields.

### DIFF
--- a/src/ai/tools.ts
+++ b/src/ai/tools.ts
@@ -1021,8 +1021,8 @@ export function createAgentTools({
 
         const visibilityNote = visible ? 'Element IS visible in browser — Tester can use this XPath as locator.' : 'Element exists in DOM but is NOT visible. May need scrolling, a click to reveal, or is hidden.';
 
-        const hit = liveElement?.attrs['data-explorbot-hit'] || null;
-        const coveredBy = liveElement?.attrs['data-explorbot-covered-by'] || null;
+        const hit = liveElement?.ourAttr('hit') || null;
+        const coveredBy = liveElement?.ourAttr('coveredBy') || null;
         let resolvedVisibilityNote = visibilityNote;
         if (hit === 'covered') {
           resolvedVisibilityNote = `Element exists in DOM and is visible, but is covered by ${coveredBy || 'another UI layer'}. Dismiss or move the covering UI before interacting with it.`;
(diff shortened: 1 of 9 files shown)


src/ai/tools.ts
```diff
@@ -1006,15 +1015,28 @@ export function createAgentTools({
 
         const action = explorer.createAction();
         const visible = await action.attempt(`I.seeElement(${JSON.stringify(xpath)})`, 'xpathCheck visibility', false);
+        const liveElement = await WebElement.fromPlaywrightLocator(explorer.playwrightHelper.page.locator(`xpath=${xpath}`));
 
         const matchesSummary = result.elements.map((el, i) => `${i + 1}. <${el.tag} ${el.keyAttrs}> text="${el.text}" html: ${el.outerHTML}`).join('\n');
 
         const visibilityNote = visible ? 'Element IS visible in browser — Tester can use this XPath as locator.' : 'Element exists in DOM but is NOT visible. May need scrolling, a click to reveal, or is hidden.';
 
+        const hit = liveElement?.attrs['data-explorbot-hit'] || null;
```
DavertMik: can we here do things less ugly
if liveEelement is instance of our webEelement class 

Add to liveElemet

```js
const hit = liveElement.ourAttr('hit')
```

src/utils/html.ts
```diff
@@ -169,14 +171,49 @@ export type ComponentScopeExtractionConfig = {
   limits: typeof HTML_EXTRACTION_LIMITS;
 };
 
+export function captureDomHtmlSnapshot(): string {
```
DavertMik: we already have like 3 functions for html snapshot:

htmlMinimalUISnapshot
htmlCombinedSnapshot
...

should we add this to thee ones?

src/utils/html.ts
```diff
@@ -246,6 +283,45 @@ export function extractElementData(el: Element, config?: ElementExtractionConfig
     return Array.from(tokens).slice(0, 8);
   }
 
+  function describeTopElement(target: Element): string {
```
DavertMik:
```suggestion
  function describeOverlappingElement(target: Element): string {
```

DavertMik (changes requested):
please simplify code

src/utils/html.ts
DenysKuchma: Upd: captureDomHtmlSnapshot -> captureHtmlForSnapshot. Its not a separate snapshot processor. It only serializes live DOM state before the existing htmlCombinedSnapshot pipeline

PR #62, opened 2026-07-01
Reuse successful test experience during execution

DenysKuchma:
## Summary
                      
  - Add experience TOC to tester context instead of full experience content. 
  - Enable tester tools to fetch relevant sections via `learnExperience`. 
  - Save fallback `FLOW` from successful steps when curator returns empty.
  - Include passed task steps when building saved experience.
  - Avoid duplicate fallback writes when relevant flow already exists. 
  - Add unit test.

### DIFF
--- a/src/ai/historian/experience.ts
+++ b/src/ai/historian/experience.ts
@@ -4,14 +4,14 @@ import { ActionResult } from '../../action-result.ts';
 import { ExperienceTracker, type SessionStep } from '../../experience-tracker.ts';
 import type { Reporter, ReporterStep } from '../../reporter.ts';
 import type { StateManager } from '../../state-manager.ts';
-import { type StepData, type Task, Test } from '../../test-plan.ts';
+import { type Task, Test } from '../../test-plan.ts';
 import { tag } from '../../utils/logger.ts';
+import { isCodeceptToolName, isNonReusableCode, mergeUniqueStepsByCode, stripComments, toReusableSessionStep } from '../../utils/step-analyzer.ts';
 import { extractStatePath } from '../../utils/url-matcher.ts';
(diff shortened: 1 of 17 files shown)


src/ai/tools.ts
```diff
@@ -20,6 +20,22 @@ const debugLog = createDebug('explorbot:tools');
 
 export const CODECEPT_TOOLS = ['click', 'hover', 'pressKey', 'form'] as const;
 export const ASSERTION_TOOLS = ['verify'] as const;
+export type CodeceptToolName = (typeof CODECEPT_TOOLS)[number];
+
+const CODECEPT_FORM_COMMANDS: readonly string[] = ['I.fillField', 'I.type', 'I.selectOption', 'I.attachFile', 'I.checkOption', 'I.uncheckOption'];
+
+function getCodeceptToolName(commandName: string): CodeceptToolName | null {
```
DavertMik: not related, find another file for this

src/ai/historian/experience.ts
```diff
@@ -96,6 +102,47 @@ export function WithExperience<T extends Constructor>(Base: T) {
...
+        tool: toolName,
+        code,
+      };
+    }
+
+    private mergeUniqueStepsByCode(primary: SessionStep[], secondary: SessionStep[]): SessionStep[] {
+      const merged: SessionStep[] = [];
+      const seen = new Set<string>();
+      for (const step of [...primary, ...secondary]) {
+        const identity = stripComments(step.code || '').trim();
```
DavertMik: i would create utils/step-analyzer.ts and have functions there to manipulate codeceptjs steps
all unit tested

as we moved all markdown manipulation into mdq 
same we can have one single place for step string manipulations

DavertMik (changes requested):

PR #78, opened 2026-07-11
fix: honor --incognito by disabling the experience tracker (plan 009)

DavertMik:
Implements **plan 009** (`plans/009-wire-incognito-flag.md`) — a P2 bug fix (deferred finding CORRECTNESS-03).

## Problem
`explorbot --incognito` promises "Run without recording experiences". The flag travels CLI → ExplorBot → Explorer, and Explorer calls `new StateManager({ incognito: ... })` — but `StateManager`'s constructor **took no parameters**, so the object argument was silently ignored and the `ExperienceTracker` it created was fully enabled. Incognito runs read and wrote `./experience/` exactly like normal runs; the user's explicit "don't learn from this session" request was dropped without any error (no typecheck gate to catch it).

## Fix
Thread `incognito` into the constructor and pass it through to the tracker, which already supports `{ disabled }`:
```ts
constructor(options: { incognito?: boolean } = {}) {
  this.experienceTracker = new ExperienceTracker({ disabled: options.incognito });
```
`disabled: undefined` collapses to `false` via the tracker's `options.disabled ?? false`, so the two argumentless `new StateManager()` call sites keep today's behavior. `explorer.ts:85` was already correct and needs no edit.

## Verification
- New tests: `new StateManager({ incognito: true })` → disabled tracker; default → enabled. (Asserted via the plan's sanctioned `(tracker as any).disabled` fallback.)
- `bun test tests/unit` → 728 pass, 0 fail; `bun run lint` clean.

## Known gap (out of scope, per plan)
Agents that construct their own `ExperienceTracker` (Navigator, Captain) bypass this flag until the DI-consolidation plan lands.

### DIFF
--- a/src/explorer.ts
+++ b/src/explorer.ts
@@ -82,8 +82,8 @@ class Explorer {
     this.aiProvider = aiProvider;
     this.options = options;
     this.initializeContainer();
-    this.stateManager = new StateManager({ incognito: this.options?.incognito });
     this.knowledgeTracker = new KnowledgeTracker();
+    this.stateManager = new StateManager({ knowledgeTracker: this.knowledgeTracker, incognito: this.options?.incognito });
     this.reporter = new Reporter(config.reporter, this.stateManager);
   }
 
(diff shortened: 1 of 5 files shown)


DavertMik:
Closing — re-creating plans 009 and 010 as separate independent PRs off main (this branch had both entangled).

PR #93, opened 2026-07-13
refactor: route all markdown parsing through mdq() (plan 022)

DavertMik:
## Plan 022 — Route all markdown manipulation through `mdq()`

The project rule (CLAUDE.md, twice) is absolute: all markdown manipulation goes through `mdq()` — never regex, `.includes()`, or line-splitting. This fixes 8 violations, including 4 hand-rolled `marked.lexer` walks that re-implemented mdq's own section logic with subtly different semantics.

### New mdq API (additive only)
`MarkdownQuery.meta()` → `Array<{type, depth, text}>` — exposes token type, heading depth, and unwrapped text. This is the gap that forced the lexer walks. 6 new unit cases in `markdown-query.test.ts`.

### Migrated
- `researcher.ts` `## Summary` regex + `researcher/parser.ts` `.includes('Extended Research')` → mdq queries.
- `planner.ts` `cleanExperienceFlows` + table swap → positional `.replace()` with **re-query-after-each-mutation** (offsets go stale after a replace). Removed all `String.replace(section.text(), …)` and `/^---/gm`.
- `experience-tracker.ts`: whole-section truncation (never cuts mid-code-fence), `writeAction`/`writeFlow` dedup via mdq, and `listTocHeadings`/`extractHeadingSection`/`renderAsHowTo` rebuilt on `meta()`.
- `experience-compactor.ts`: `listSections` on `section2`+`meta()`; the re-hardcoded clickXY regex → `isNonReusableCode()`.
- `knowledge-tracker.ts` firstLine → `mdq().meta()`.
- **`marked` import removed** from experience-tracker and experience-compactor.

### Two deviations (both improvements, verified)
1. **`renderAsHowTo` uses `startsWith` via `meta()`**, not the plan's literal `h2(~"marker")`. That contains-matcher would infinite-loop on a title containing the marker (`## FLOW: undo FLOW: steps`) and diverges from the old `startsWith`. The implemented version provably terminates and matches the old output (whitespace-only diff — the old code accidentally doubled blank lines).
2. **`cleanExperienceFlows`' `> ... and N more discoveries` note was dead code** — the old `.replace()` no-op'd because the tail blockquotes were already removed from the string. The new code correctly appends it, as the plan specifies.

### STOP-candidates cleared
- **Step 5 mixed-depth section indexing** — verified **byte-identical** to the old lexer walk across all indices (h1–h4, out-of-range).
- **Step 7 clickXY → `isNonReusableCode` broadening** — no test pins the narrow behavior; the `#ember123` case yields the same final result (dropped at strip stage). All 43 compactor tests pass.

### Verification
- `bun test tests/unit` → **772 pass / 0 fail** (+8 new tests; no existing assertions modified).
- `bun test tests/integration/` → **63 pass / 0 fail**.
- `bun run lint` → clean; `tsc` total neutral (639 = 639).
- `grep marked` in both files → empty; `## Summary` regex gone; `meta()` present.

### DIFF
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -280,29 +280,43 @@ export class Planner extends PlannerBase implements Agent {
   }
 
   private cleanExperienceFlows(text: string): string | null {
-    const seenTitles = new Set<string>();
     let result = text;
 
-    for (const section of [...mdq(text).query('section2').each(), ...mdq(text).query('section3').each()]) {
-      const heading = section.query('heading').text().trim();
-      const body = mdq(section.text())
-        .query('heading')
-        .replace('')
-        .replace(/^---\s*$/gm, '')
-        .trim();
-
-      if (!body || seenTitles.has(heading)) {
-        result = result.replace(section.text(), '');
-        continue;
+    while (true) {
+      const sections = [...mdq(result).query('section2').each(), ...mdq(result).query('section3').each()];
+      const seenTitles = new Set<string>();
+      let removed = false;
+      for (const section of sections) {
+        const heading = section.query('heading').text().trim();
+        const withoutHeadings = mdq(section.text()).query('heading').replace('');
+        const body = mdq(withoutHeadings).query('hr').replace('').trim();
+        if (body && !seenTitles.has(heading)) {
+          seenTitles.add(heading);
+          continue;
+        }
+        result = section.replace('');
+        removed = true;
+        break;
       }
-      seenTitles.add(heading);
+      if (!removed) break;
+    }
 
-      const blockquotes = section.query('blockquote').each();
-      if (blockquotes.length <= 10) continue;
-      for (const bq of blockquotes.slice(10)) {
-        result = result.replace(bq.text(), '');
+    const trimmedTitles = new Set<string>();
+    while (true) {
+      const sections = [...mdq(result).query('section2').each(), ...mdq(result).query('section3').each()];
+      let trimmed = false;
+      for (const section of sections) {
+        const heading = section.query('heading').text().trim();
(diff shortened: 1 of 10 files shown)


DenysKuchma (approved):

src/ai/planner.ts
```diff
@@ -280,29 +280,43 @@ export class Planner extends PlannerBase implements Agent {
...
-      seenTitles.add(heading);
+      if (!removed) break;
+    }
 
-      const blockquotes = section.query('blockquote').each();
-      if (blockquotes.length <= 10) continue;
-      for (const bq of blockquotes.slice(10)) {
-        result = result.replace(bq.text(), '');
+    const trimmedTitles = new Set<string>();
+    while (true) {
```
DavertMik: do not do while true!!!!
seems like we should simplify it

src/experience-tracker.ts
```diff
@@ -535,21 +522,18 @@ function generateActionContent(title: string, code: string, explanation?: string
...
-      if (text.startsWith('ACTION:')) {
-        result += `## HOW to ${text.slice(7).trim()} (single-step)\n\n`;
-        continue;
-      }
+  let result = content;
+  for (const { marker, suffix } of [
+    { marker: 'FLOW:', suffix: 'multi-step' },
+    { marker: 'ACTION:', suffix: 'single-step' },
+  ]) {
+    while (true) {
```
DavertMik: model keeps doing while(true)
i think we can move this iterator to the mdq

like 
```js
mdq(result).forEach('h2', ...)
```

DavertMik (commented):
Idea is good but while true must be avoided
it doesn't simplify things!

PR #106, opened 2026-08-03
Prima boat — high-level browser driver for orchestrating agents

DavertMik:
## What

**prima** (`boat/prima`): an intent-level browser driver CLI that lets an orchestrating agent (Claude Code / Opus) drive a browser through explorbot's cheap-model perception layer. Spec + plan + full implementation (11 tasks, per-task adversarial reviews, whole-branch final review).

```
explorbot prima pw "({ page }) => page.click('text=Login')"   # precise, no AI
explorbot prima do "open the first invoice" "download its PDF" # tester-style NL
explorbot prima ask "what do I see?"                           # vision Q&A
explorbot prima research --data                                # verified UI map
explorbot prima verify "user is logged in"                     # AI assertion
explorbot prima go https://app.example.com                     # navigate
prima browser start|stop|status|list                           # owned instances
```

Every command returns a uniform envelope (result + `used:` executed code, page/state, ariaDiff, instance info, artifact file paths) — page data never enters the orchestrator's context. Failures heal via Navigator's cheap-model recovery with an attempt trace; exhaustion returns compact ARIA inline + full HTML/ARIA/network as files.

## Highlights

- **Attaches to playwright-cli's browser by default** via the Playwright-protocol registry (`~/.cache/ms-playwright/b/`, matched by workspaceDir + session title, probe-before-select); prima-owned named instances as fallback; **no implicit browser launch ever** (fails with create-a-session guidance)
- **AI-optional**: `pw` and URL-`go` work with no model (`optionalAi` bootstrap); NL commands fail cleanly suggesting playwright-cli fallback
- **Config-free operation**: `~/.explorbot/` global config + `.env`, per-host state dirs (`~/.explorbot/state/<host>/`) with experience memory; `--ephemeral` for throwaway runs
- **Core additions**: named browser-server instances, Navigator `onAttempt` hook, verbatim execution path for the generated pw wrapper, `~/.explorbot` config ladder
- **Ships to Node**: `npx explorbot prima ...` proven end-to-end (browser start → go → pw, no stubs); `prima` is a second bin of the package. Bun-from-source covers everything except Playwright `connect()` flows (documented caveat)
- 1055 tests pass (unit + aimock integration + real-browser smoke); docs + changelog included

## Found & fixed along the way

- `usePlaywrightTo` rejects non-async fns — every documented `pw` form failed (smoke caught it)
- Two undeclared `playwright-core@1.54` imports causing protocol skew vs `playwright@1.60` (one made the Node build's connect fail)
- `browser start` didn't hold Bun's event loop (latent in main CLI too)
- Config-load ordering + throwing failure-envelope path in Prima.start()

## Review fixes

- `go <absolute url>` double-navigation — fixed in 5078118. The url-shaped target is seeded into a dedicated `PrimaOptions.baseUrl`, which only feeds config resolution, so `options.url` keeps meaning "open this page at start". `go` navigates once, through the heal layer.
- Per-host state dir could be named `state/localhost:3000`, which Windows cannot create — the colon is replaced, so hosts differing only by port stay apart and IPv6 literals work.
- A file materialized from `EXPLORBOT_KNOWLEDGE_FILE` outlived its run now that config-free output is persistent — it goes to a machine-owned `knowledge/env/` directory cleared on every load, so no stale knowledge or credentials reach a later run. Hand-authored knowledge files are untouched.
- Lint is green (`noDelete` on a test env unset).

## Follow-ups (not in this PR)

- Investigate `connectOverCDP` under Bun to close the connect() gap
- Live playwright-cli (≥1.62) attach smoke before release tag — settles adopted-context disconnect semantics
- Declare `playwright-core` pinned to `playwright`'s version (playwright-recorder still resolves the hoisted copy)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -709,7 +709,7 @@ export function resolveStateRoot(baseUrl: string, ephemeral?: boolean): string {
   const host = URL.parse(baseUrl)?.host;
   if (ephemeral || !host) return mkdtempSync(join(tmpdir(), 'explorbot-'));
 
-  const stateRoot = join(globalDir(), 'state', host);
+  const stateRoot = join(globalDir(), 'state', host.replaceAll(':', '_'));
   mkdirSync(stateRoot, { recursive: true, mode: 0o700 });
   return stateRoot;
 }
@@ -723,8 +723,10 @@ export function materializeKnowledge(outputRoot: string): void {
   const knowledgeFile = process.env.EXPLORBOT_KNOWLEDGE_FILE;
   const knowledgeDir = join(outputRoot, 'knowledge');
   const globalFile = join(knowledgeDir, 'global.md');
+  const envDir = join(knowledgeDir, 'env');
 
   if (!inline) rmSync(globalFile, { force: true });
+  rmSync(envDir, { recursive: true, force: true });
   if (!inline && !knowledgeFile) return;
 
   mkdirSync(knowledgeDir, { recursive: true });
@@ -739,7 +741,8 @@ export function materializeKnowledge(outputRoot: string): void {
   if (!existsSync(source)) {
     throw new Error(`Knowledge file from EXPLORBOT_KNOWLEDGE_FILE not found: ${source}`);
   }
-  copyFileSync(source, join(knowledgeDir, basename(source)));
+  mkdirSync(envDir, { recursive: true });
+  copyFileSync(source, join(envDir, basename(source)));
 }
 
 export async function createModel(provider: string, modelId: string): Promise<any> {
--- a/src/config.ts
+++ b/src/config.ts
@@ -542,7 +542,7 @@ export class ConfigParser {
       playwright: { browser: 'chromium', url, show: false },
       ai,
       dirs: { knowledge: 'knowledge', experience: 'experience', output: '.' },
-      experience: { disabled: true },
+      experience: { disabled: !!process.env.EXPLORBOT_EPHEMERAL },
     };
   }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -705,10 +705,11 @@ export function resolveOutputRoot(baseUrl?: string): string {
 }
 
 export function resolveStateRoot(baseUrl: string, ephemeral?: boolean): string {
-  if (ephemeral) return mkdtempSync(join(tmpdir(), 'explorbot-'));
+  const host = URL.parse(baseUrl)?.host;
+  if (ephemeral || !host) return mkdtempSync(join(tmpdir(), 'explorbot-'));
 
-  const stateRoot = join(globalDir(), 'state', new URL(baseUrl).host);
-  mkdirSync(stateRoot, { recursive: true });
+  const stateRoot = join(globalDir(), 'state', host);
+  mkdirSync(stateRoot, { recursive: true, mode: 0o700 });
   return stateRoot;
 }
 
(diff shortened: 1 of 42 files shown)


src/config.ts
```diff
@@ -664,30 +685,51 @@ export async function resolveModel(spec: string, role: ModelRole = 'model'): Pro
...
+    cachedOutputRoot = resolveStateRoot(baseUrl, !!process.env.EXPLORBOT_EPHEMERAL);
+    return cachedOutputRoot;
+  }
+
+  cachedOutputRoot = mkdtempSync(join(tmpdir(), 'explorbot-'));
   return cachedOutputRoot;
 }
 
+export function resolveStateRoot(baseUrl: string, ephemeral?: boolean): string {
+  const host = URL.parse(baseUrl)?.host;
```
DenysKuchma, line 709: The URL host is used directly as a directory name. For URLs with a port this produces paths such as state/localhost:3000, which cannot be created on Windows

src/config.ts
```diff
@@ -664,30 +685,51 @@ export async function resolveModel(spec: string, role: ModelRole = 'model'): Pro
...
+  return join(os.homedir(), '.explorbot');
+}
+
 export function materializeKnowledge(outputRoot: string): void {
   const inline = process.env.EXPLORBOT_KNOWLEDGE;
   const knowledgeFile = process.env.EXPLORBOT_KNOWLEDGE_FILE;
+  const knowledgeDir = join(outputRoot, 'knowledge');
+  const globalFile = join(knowledgeDir, 'global.md');
+
+  if (!inline) rmSync(globalFile, { force: true });
```
DenysKuchma, line 728: Config-free output is now persistent, but files materialized through EXPLORBOT_KNOWLEDGE_FILE are never removed when the variable is absent or points to another file on the next run. Stale knowledge, including credentials, can therefore leak into later runs for the same host

DenysKuchma (changes requested):

DenysKuchma:
Also fix Regression
Also fix Lint

DenysKuchma:
Also must be resolve double navigation prima go <absolute-url>

src/config.ts
DavertMik: Fixed in cf967f9 — the colon is replaced in the directory name, so a port lands as `state/localhost_3000` (creatable on Windows, and IPv6 literals work too). Hosts that differ only by port still get separate dirs; covered by two tests in `tests/unit/config-ladder.test.ts`.

src/config.ts
DavertMik: Fixed in cf967f9 — `EXPLORBOT_KNOWLEDGE_FILE` now materializes into a machine-owned `knowledge/env/` directory that is removed on every load, so a run that drops the variable or points it at another file cannot inherit the previous one. Hand-authored knowledge files in the same directory are left alone; `tests/unit/config-ladder.test.ts` covers both the swap and the removal.

DavertMik:
Both review comments are addressed in cf967f9, and `origin/main` is merged in (the branch no longer conflicts).

**Lint** — green. The failure was `lint/performance/noDelete` on `delete process.env.EXPLORBOT_KNOWLEDGE` in `tests/unit/config-ladder.test.ts`; the unset now goes through a named key, like every other env unset in that file. `test (24)` passes on the new head.

**`prima go <absolute-url>` double navigation** — this landed in 5078118, before the review; the PR description still carried it as an open residual, which I have now corrected. `options.url` means "open this page at start", so seeding a url-shaped `go` target into it made the target load twice and let a failing start-visit bypass `go`'s heal path. The target is now seeded into a dedicated `PrimaOptions.baseUrl` that only feeds config resolution. Two tests in `boat/prima/tests/prima.test.ts` pin it: one asserts a base url reaches the configuration and never the browser (`go:` is the only navigation recorded), the other that an explicit `--url` is still opened when the session has no page.

**Regression** — the run is queued on the new head and waiting at the `regression` environment gate for your approval; I cannot approve it. The previous failure was on the "tests passed" criterion in all three attempts, with the session analysis attributing it to the fixture's filter behaviour (search filter returns nothing, filter types cannot be combined) rather than to a navigation or recovery failure — login, research, and scenario generation passed every attempt. The same scenario passed on this branch on 2026-08-03 and has failed and passed on other branches in the same window, so I read it as scenario-selection variance, not a break from this PR. Happy to dig into the fixture's filtering if the re-run fails the same way.

PR #108, opened 2026-08-04
Reuse Docbot application specs as Explorbot context

DenysKuchma:
## Summary

  Adds support for reusing documentation generated by Docbot as additional context in Explorbot.

  Docbot page files now include a versioned application-spec contract. Explorbot can load the generated bundle from configuration or through the `--spec` CLI option, match documentation to the current URL, and provide it to Researcher, Planner, Navigator, Tester, and other agents.

  ## Changes

  - add `applicationSpec` configuration option
  - add `--spec <path>` to web CLI commands
  - introduce a validated, versioned application-spec format
  - load only documentation relevant to the current page
  - keep live HTML, ARIA, and screenshots as the primary source of truth
  - treat `User Might` capabilities as unverified context
  - separate application-spec context from regular knowledge
  - support Docbot bundle directories and `index.md` paths
  - tolerate SPA-added URL fragments during navigation
  - document the Docbot-to-Explorbot workflow

### DIFF
--- a/src/application-spec-contract.ts
+++ b/src/application-spec-contract.ts
@@ -1,2 +1,10 @@
+import { z } from 'zod';
+
 export const APPLICATION_SPEC_FORMAT = 'explorbot-application-spec';
 export const APPLICATION_SPEC_VERSION = 1;
+
+export const APPLICATION_SPEC_PAGE_SCHEMA = z.object({
+  format: z.literal(APPLICATION_SPEC_FORMAT),
+  version: z.literal(APPLICATION_SPEC_VERSION),
+  url: z.string().trim().min(1),
+});
--- a/src/application-spec-contract.ts
+++ b/src/application-spec-contract.ts
@@ -1,0 +1,2 @@
+export const APPLICATION_SPEC_FORMAT = 'explorbot-application-spec';
+export const APPLICATION_SPEC_VERSION = 1;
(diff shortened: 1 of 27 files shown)


src/knowledge-tracker.ts
```diff
@@ -24,15 +25,22 @@ export class KnowledgeTracker {
   private knowledgeDir: string;
   private knowledgeFiles: Knowledge[] = [];
   private isLoaded = false;
+  private applicationSpec?: ApplicationSpec;
 
-  constructor() {
+  constructor(applicationSpecPath?: string) {
```
DavertMik, line 30: should be loaded same as knowledge from config just as  

```js
this.specDir = configParser.resolveProjectDir(config.dirs?.spec);
```

src/application-spec-contract.ts
```diff
@@ -0,0 +1,2 @@
+export const APPLICATION_SPEC_FORMAT = 'explorbot-application-spec';
+export const APPLICATION_SPEC_VERSION = 1;
```
DavertMik, line 4: I expected contract will actually write how spec must be written and structured

DavertMik (changes requested):
We also need document what application SPEC is into docs/ so non-docbot specs could also be accepted

DavertMik (approved):

PR #109, opened 2026-08-06
Fix what the tester is told: dead vision tools, doomed clicks, and misrouted click failures

DavertMik:
Five fixes from a Langfuse review of the 2026-08-06 CI session (6 runs, 84 tests, 151 traces, 14,626 observations, 871 click calls of which 463 failed).

| | |
|---|---|
| Tests attempted | 84 |
| Reached `finish` | 47 (7 of which *report failure*) |
| Called `stop` | 9 |
| Never ended (max iterations/abort) | 28 |

## 1. Vision tools stayed callable after the vision model died

The configured vision model returned `does not exist or you do not have access to it` on all 192 calls — half of every error in the session. Explorbot already degraded correctly (set `visionDisabled`, told the AI to use ARIA), but `see` and `visualClick` stayed in the toolset and kept answering. The AI called them **230 times and got an error back on 220** — steps burned against the iteration cap, on precisely the fallback it reaches for when a click fails.

`disableVision()` now withdraws both tools. Consumers spread the record fresh each iteration, so they vanish from the next prompt. In-tool guards stay for calls already emitted in the same roundtrip; `pilot.pickPlanningTools()` already guarded with `if (see)`.

## 2. A doomed click blocked for 30 seconds

Page interactions inherited `playwright.timeout`, which reaches `page.setDefaultTimeout()`. At Playwright's 30s default, clicking a disabled or unreachable control held the test for the full 30s. **137 such timeouts in one session — roughly 69 minutes, about a third of the 3.5h wall clock.**

New `action.timeout` (default `3000`) applied to the page before each action. Measured against a real Chromium and a disabled button with `playwright.timeout: 30000`: **30015ms → 3004ms**.

## 3. 187 of 463 click failures were told the wrong cause

Two misroutes, both deterministic code behavior rather than statistics:

- A container miss throws `Clickable element X was not found inside element Y`. That contains `was not found`, so it was reported as **"not found in the DOM"** — but the element usually exists; the *container* was wrong. 117 cases.
- A disabled control surfaces as a Playwright timeout, so it was reported as **"covered by overlay… or use visualClick()"** — no overlay is involved, a precondition is unmet, and `visualClick` cannot click a disabled button either. 46 cases.

Downstream cost: after a disabled failure the model went hunting for a different locator **77%** of the time, and those tests recovered least often (**54%** vs 69% for genuine locator misses). One chain re-clicked the same disabled `Select` three times, then abandoned the test — and that is one of the 7 false `finish` calls above.

`clickFailureSuggestion()` classifies across **all** attempts rather than `attempts[last]` (with a fallback ladder the last error is usually the least informative) and orders existence evidence ahead of not-found: disabled → covered → hidden → container miss → absent. Replayed over the session's real failures it classifies **346/346** correctly. Covered by 6 new unit tests pinning the real CodeceptJS/Playwright error strings, since those are the drift risk on upgrade.

## 4. The rules taught an ARIA locator that can never match

`role: "input"` was listed under `<good_aria_locator_example>`. It is not an ARIA role. Verified against a real browser: `getByRole('input')` returns **0** elements for a text input, `getByRole('textbox')` returns 1. Same mistake in a CSS example (`div[role=input]`).

ARIA stays the first choice — generality and the maintainability of the generated CodeceptJS suites depend on it. What changed is that role and text must be copied from the ARIA snapshot or UI map rather than guessed, which is where the invented pairs came from. Single-command clicks using `role="button"` succeeded only 29% (20/68) against `role="link"` at 54%.

## 5. Container guidance contradicted itself

Two rules said containers were the default for every interaction while a third reserved them for disambiguation. Now conditional and consistent across all three places, and the containerless fallback is a stated requirement rather than a footnote — 117 failures exhausted their fallback ladder still holding a container.

Deliberately **not** claiming a performance win here. A paired test (same target locator tried both ways within the same test) gives 11-5 on 16 discordant pairs, p = 0.21, and no difference in 51 of 67 — the raw 56%-vs-42% gap is mostly selection bias, since the model reaches for containers on harder targets. This change removes a contradiction, nothing more.

## What was deliberately not done

Flipping locator priority toward CSS. CSS scores higher (59% vs 42% on single-command calls) but the comparison is confounded the same way, and CSS-first would make the generated suites brittle and app-specific — the reason ARIA is preferred in the first place. Worth re-measuring in a paired comparison once invented pairs drop; if ARIA still lags, the fix belongs in the snapshot pipeline, not the priority ladder.

## Verification

- `bun test tests/unit` — 847 pass / 2 fail / 2 errors; the 2 failures are identical on `origin/main` and pre-existing
- `bun test tests/integration/` — 79 pass, 0 fail
- Timeout behavior and the `role: "input"` claim both measured end-to-end against a real Chromium
- New classifier replayed against every real failing click in the session

`tests/unit/explorer.test.ts` gains `setDefaultTimeout` on its mock page — the mock was missing a method the real Playwright `Page` always has.


src/action.ts
```diff
@@ -297,6 +298,8 @@ class Action {
         throw new Error('No valid I.* or page.* commands found in code block');
       }
 
+      this.playwrightHelper?.page?.setDefaultTimeout(this.config.action?.timeout ?? DEFAULT_ACTION_TIMEOUT);
```
DenysKuchma, line 306: setDefaultTimeout() mutates the Page permanently. After the first action, the 3-second action timeout also applies to later navigation, waits, and direct Playwright operations, effectively overriding playwright.timeout for the rest of the session

src/ai/tools.ts
```diff
@@ -1011,6 +1002,13 @@ export function createAgentTools({ explorer, stateManager, ai, researcher, navig
     }),
   };
 
+  const disableVision = (): void => {
+    visionDisabled = true;
+    Reflect.deleteProperty(tools, 'see');
+    Reflect.deleteProperty(tools, 'visualClick');
```
DenysKuchma: Captain keeps copies of these tools, so deleting them here has no effect there
Rebuild or filter Captain tool list on each iteration

DenysKuchma (changes requested):

src/action.ts
DavertMik: Fixed in 85c976a. The page timeout is now restored to `playwright.timeout` as soon as the commands finish, and again in the `finally` so the throw path is covered too — same set/restore shape `Navigator.verify()` already uses. The short budget now covers the interaction only; navigation, waits, and page capture run under the configured Playwright timeout again.

src/ai/tools.ts
DavertMik: Fixed in 85c976a. You are right, and it was worse than the copies: `visionDisabled` was a closure variable per `createAgentTools()` call, so Pilot, Tester, Rerunner and every Captain `handle()` each had their own — one agent losing vision left the others calling a dead model.

The flag moved to `Stats.visionDisabled` (session-wide), and `withdrawVisionTools(tools)` applies it in three places: at the end of `createAgentTools()` so any toolset built later omits them, in `disableVision()` for records already handed out, and at the top of each Captain loop iteration for the record built before the model died. `web-mode.ts` now attaches `see`/`visualClick` only when they exist, following the `pilot.pickPlanningTools()` pattern, so a withdrawn tool never lands in the record as `undefined`.

Covered by three tests in `tests/unit/agent-tools.test.ts`.

DavertMik: Follow-up in bd8971b: the `Stats` flag alone only covered toolsets built *after* the model failed. Pilot and Rerunner build theirs once at construction and Tester spreads a cached record, so a failure inside one of them still left the others offering `see`. All three now filter the record they hand to the model, the same way Captain does.

PR #119, opened 2026-08-17
Prima verdict honesty: no reload in check, vision-confirmed outcomes, unconfirmed is not failed

DavertMik:
A field run of prima found the verdicts untrustworthy in both directions: `do` reported failures for clicks that had landed, and `check` returned `ok: false` for an environmental reason without saying what it was. A false red trains the caller out of trusting `### Result`, and then the greens stop meaning anything either.

## `check` was reloading the page it was asked about

`prima.check()` → `tester.test()` → `runTestSession` → `explorer.visit(task.startUrl!)` (`tester.ts:192`, unconditional) → `I.amOnPage()` (`explorer.ts:431`) → `page.goto()`, which reloads even when the address is unchanged. `task.startUrl` is the page the caller is already on, so an open dialog, a selected tab or a filled form was destroyed before anything was checked. This was guaranteed, not a race with a dev-server reload.

`Tester.test(task, opts)` gained `startOnCurrentPage`, which skips the initial visit; prima passes it. The explore flow is untouched, where reload-to-start-url is intentional. `reset` needed no guard — it already refuses when the current URL equals the start URL.

## `check` now confirms its verdict by looking at the page

Outcomes were settled from a text log, so a claim about layout or position — anything an accessibility tree cannot express — was judged by something that had never seen the page. `settleExpectations` (called only from prima) now takes the final `ActionResult` and judges every outcome against the run log **and** a screenshot of the final page, in one structured call on the vision model.

It judges *all* outcomes rather than only undecided ones, so a DOM assertion the run already made can still be contradicted by the page. The prompt states the screenshot shows the final page only, so an outcome established earlier stays established.

`### Expected outcomes` ends with a `judged from` line naming the evidence the verdict actually had. When the vision call cannot produce a verdict it falls back to the text model and flips `Stats.visionDisabled`, so that line reports the log rather than a confirmation that never happened — without it, deriving `ok` from the outcomes would turn a crashed judge into a green verdict claiming a screenshot.

## `check`'s verdict is its outcomes

`ok` came from `tester.test()`'s success flag and could contradict the outcomes printed beside it. It now follows them: `ok: true` when nothing FAILED. An outcome the run never checked stays "not verified" and is not a failure. A run that could not complete is reported as such rather than as an application failure.

## `do` separates failed from unconfirmed

`do`'s verdict was the model's bookkeeping, not the page. An instruction never passed to `completed()` rendered as `error: open: <instruction>`, so an envelope could show every step green and `ok: false` at once. Unreported instructions are now `??` rows in `### Steps` — the actions that ran are listed above them — and do not fail the command. Only action errors and `blocked()` do. An AI error while settling the ledger is reported as its own step instead of being blamed on the instruction.

## Also

- `verify` judges a claim from a screenshot when no assertion can express it, instead of dead-ending on a suggestion nothing acts on. Tester's `verify` tool calls the same `navigator.verifyState`, so this covers `check` too.
- `status <hash>` returns the page block and artifact paths; the compact ARIA tree is no longer stored in `status.json` and reprinted inline.
- `prima <command> --help` documents the `??` marker and the `judged from` line — a marker visible in an envelope but nowhere lookupable is not documented.

## Verification

`bun test tests/unit tests/integration boat/prima/tests` — 1198 pass. The one failure, `remote > announces the run…`, fails identically on a clean tree in full-suite runs and passes in isolation; it is pre-existing and flaky. No new type errors.

New coverage: `tests/unit/pilot-settle-vision.test.ts` (vision judge, its fallback, and the no-screenshot path) plus eight cases in `boat/prima/tests/prima.test.ts` for the verdict changes.

Design recorded in `docs/superpowers/specs/2026-08-18-prima-verdict-honesty.md`.

## Not done

- `explorer.beginTest` still calls `closeOtherTabs()`, so `check` closes other tabs of an attached session. Guarding it means threading an option through `beginTest`, which every flow shares.
- `do` gets no vision confirmation pass; a per-instruction vision call is a different cost profile.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,44 @@
 # Changelog
 
+## 2026-08-18
+
+### `prima check` no longer reloads the page it was asked about
+
+`check` used to navigate to the current URL before it started, which reloads the page even when the
+address is unchanged. Anything the page was holding — an open dialog, a selected tab, a filled but
+unsaved form — was gone before the first thing was checked. It now runs on the page already open.
+
+### A `check` verdict is confirmed by looking at the page
+
+Outcomes used to be settled from the run log alone, so a claim about layout, position or anything
+else the page structure cannot express was judged by something that had never seen it. Each outcome
+is now judged against the run log **and** a screenshot of the page as the run left it. The
+`### Expected outcomes` section ends with a `judged from` line naming which of the two the verdict
+actually had, so "the run log alone" — no vision model configured, or the vision model could not
+answer — is visible rather than silent.
+
+`ok:` now follows those outcomes: false only when one FAILED. An outcome the run never checked
+stays "not verified" and does not fail the command. A run that could not complete at all says so,
+instead of reporting it as a failure of the application.
+
+### Changes
+
+- `prima do` marks an instruction it could not confirm `??` in `### Steps` instead of reporting it
+  as an error. Previously an instruction whose action had landed, but which the model never got
+  round to reporting, failed the whole command — a red for missing paperwork. The actions that ran
+  are listed above the `??` row. Only a failed action and an instruction the page could not carry
+  out fail the command now.
+- `prima do` reports an AI error while it was working out which instructions were satisfied as its
+  own step, rather than blaming the instruction that went unreported because of it.
+- `prima verify` judges a claim from a screenshot when no assertion can express it, instead of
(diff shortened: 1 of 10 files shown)


DavertMik:
Superseded by #120 — same commit, renamed to say what it fixes.

PR #120, opened 2026-08-17
Fix prima's false verdicts: check reloads the page it inspects, unreported instructions count as errors

DavertMik:
A field run found prima returning verdicts that did not match what happened, in both directions: `do` reported failures for clicks that had landed, and `check` returned `ok: false` for an environmental reason without saying what it was.

## `check` was reloading the page it was asked about

`prima.check()` → `tester.test()` → `runTestSession` → `explorer.visit(task.startUrl!)` (`tester.ts:192`, unconditional) → `I.amOnPage()` (`explorer.ts:431`) → `page.goto()`, which reloads even when the address is unchanged. `task.startUrl` is the page the caller is already on, so an open dialog, a selected tab or a filled form was destroyed before anything was checked. This was guaranteed, not a race with a dev-server reload.

`Tester.test(task, opts)` gained `startOnCurrentPage`, which skips the initial visit; prima passes it. The explore flow is untouched, where reload-to-start-url is intentional. `reset` needed no guard — it already refuses when the current URL equals the start URL.

## `check` settles its outcomes against a screenshot, and reports a disagreement

Outcomes were settled from a text log, so a claim about layout or position — anything an accessibility tree cannot express — was judged by something that had never seen the page. `settleExpectations` (called only from prima) now takes the final `ActionResult` and settles every outcome in one structured call on the vision model with the screenshot attached.

The screenshot is the proof, not one of two equal inputs: an outcome is satisfied when the page shows it to somebody looking at it, and the run log only says what the run did.

**Where the two disagree, the judge does not choose.** It returns `contradiction` with what each side shows, and hands over the files to settle it on:

```
### Expected outcomes
1. PASSED       the Add connection dialog opens
2. CONTRADICTION the new connection is listed
      the assertion I.see("staging-db") found the row in the page structure; the screenshot
      shows the list still empty, so the row is in the DOM but not visible
3. not verified  the list scrolls

### Artifacts
aria:       output/prima/9f2c1ab/aria.yml
html:       output/prima/9f2c1ab/page.html
screenshot: output/prima/9f2c1ab/page.png
```

A contradiction is the one verdict the caller cannot act on from the envelope alone, so it carries the page's html, aria and screenshot on disk — a stronger model opens them and settles it rather than taking the judge's word. The word is `contradiction` rather than `conflict` because it names what happened, and rather than `ambiguity` because that is what `unverified` already means.

A value present in the markup but absent from the picture — hidden, collapsed, covered by an overlay, drawn off-screen — is a defect in the application, and it is exactly what the other two verdicts destroy: `passed` buries it behind an assertion that happens to match, `failed` mislabels a feature that half works. A contradiction fails the command.

The prompt keeps the guard that the screenshot is the final page only, so an outcome established earlier and since navigated past is not reported as a conflict.

There is no line saying which evidence the judge held — that is plumbing, not a fact about the app. Only the degraded case is stated, as a `### Warning` when no screenshot backed the outcomes: no vision model configured, or the vision call could not produce a verdict, which falls back to the text model and flips `Stats.visionDisabled`.

## Answering the review on off-screen elements

Review raised that off-screen elements would produce false conflicts. Captures here are already `fullPage: true` (`src/action.ts:107`), so the judge sees the whole scrollable document rather than a viewport crop — but the prompt caused the predicted bug anyway, by treating "absent from the picture" as a contradiction.

The fix is in the prompt, not in code:

- **A conflict requires a positive contradiction.** The picture must show something incompatible — a list visibly empty, an error where a result was expected, the old value still displayed. An element merely not made out is `unverified`, which does not fail the command. Absence of evidence is not evidence of absence.
- **The picture does not reach inside a region that scrolls on its own.** `fullPage` expands to *document* height, so an `overflow: auto` container still clips its content. The prompt says so, so a row below an inner fold is not called a conflict.

An earlier revision of this branch probed the DOM to explain *why* something was invisible — colour against backdrop, `elementFromPoint` hit-testing, scroll extents. It was 150 lines of heuristics answering a question nobody asked, and several of them would be wrong routinely. It is removed in 871a0da. "The run says it is there and the picture does not show it" is the finding; explaining why is the caller's job, on a page they can open.

## `check`'s verdict is its outcomes

`ok` came from `tester.test()`'s success flag and could contradict the outcomes printed beside it. It now follows them: `ok: true` when nothing FAILED and nothing CONTRADICTIONed. An outcome the run never checked stays "not verified" and is not a failure. A run that could not complete is reported as such rather than as an application failure.

## `do` separates failed from unconfirmed

`do`'s verdict came from the model's bookkeeping rather than the page. An instruction never passed to `completed()` rendered as `error: open: <instruction>`, so an envelope could show every step green and `ok: false` at once. Unreported instructions are now `??` rows in `### Steps` — the actions that ran are listed above them — and do not fail the command. Only action errors and `blocked()` do. An AI error while settling the ledger is reported as its own step instead of being blamed on the instruction.

## Also

- `verify` judges a claim from a screenshot when no assertion can express it, instead of dead-ending on a suggestion nothing acts on. Tester's `verify` tool calls the same `navigator.verifyState`, so this covers `check` too.
- `status <hash>` returns the page block and artifact paths; the compact ARIA tree is no longer stored in `status.json` and reprinted inline.
- `prima <command> --help` documents the `??` marker and the `CONTRADICTION` status — a marker visible in an envelope but nowhere lookupable is not documented.

## Verification

`bun test tests/unit tests/integration boat/prima/tests` — 1198 pass. The one failure, `remote > announces the run…`, fails identically on a clean tree in full-suite runs and passes in isolation; it is pre-existing and flaky. No new type errors.

New coverage: `tests/unit/pilot-settle-vision.test.ts` (vision judge, its fallback, and the no-screenshot path) plus eight cases in `boat/prima/tests/prima.test.ts` for the verdict changes.

Design recorded in `docs/superpowers/specs/2026-08-18-prima-false-verdicts.md`.

## Not done

- `explorer.beginTest` still calls `closeOtherTabs()`, so `check` closes other tabs of an attached session. Guarding it means threading an option through `beginTest`, which every flow shares.
- `do` gets no vision confirmation pass; a per-instruction vision call is a different cost profile.


src/ai/pilot.ts
```diff
@@ -562,24 +563,44 @@ export class Pilot implements Agent {
...
 
+    let pageEvidence = '';
+    if (image) {
+      pageEvidence = dedent`
+        A screenshot of the page as the run left it is attached. It is the proof: an outcome is satisfied when
+        the page shows it to somebody looking at it. The log only says what the run did.
+
+        Where the two disagree, do not choose between them. Report "conflict" and say what each side shows.
+        Something present in the page structure but absent from the picture is a finding about the application —
+        it is not a verdict to settle.
```
DenysKuchma: A screenshot only covers the viewport. Off-screen elements may incorrectly cause a conflict

src/ai/pilot.ts
```diff
@@ -562,24 +563,44 @@ export class Pilot implements Agent {
...
+    if (image) undecided = task.expected;
     if (!undecided.length) return task.expected.map((text) => ({ text, status: decided(text) }));
 
     const schema = z.object({
       outcomes: z.array(
         z.object({
           expectation: z.string().describe('The expected outcome, repeated exactly as it was given'),
-          status: z.enum(['passed', 'failed', 'unverified']).describe('passed = the log shows it happened, failed = the log shows it did not, unverified = the run never established either way'),
+          status: z.enum(['passed', 'failed', 'unverified', 'conflict']).describe('passed = the evidence shows it happened, failed = the evidence shows it did not, unverified = the run never established either way, conflict = the page and the run disagree'),
+          evidence: z.string().nullable().describe('What settled it. For a conflict, what each side shows. Null when there is nothing to add'),
```
DenysKuchma: A screenshot only covers the viewport. An off-screen element could be incorrectly reported as a conflict

DenysKuchma (changes requested):

src/ai/pilot.ts
DavertMik: Good catch on the outcome, though the premise is different here: captures are already `fullPage: true` (`src/action.ts:107`), and `ActionResult.screenshot` loads that file, so the judge sees the whole scrollable document rather than a viewport crop.

The bug you predicted was real anyway, and it was this prompt's fault. Saying "present in the page structure but absent from the picture is a finding" turns every element the model cannot make out into a conflict. Fixed two ways in 7dc656e:

1. **A conflict now needs a positive contradiction** — the picture showing a list visibly empty, an error where a result was expected, the old value still displayed. "I cannot find it" is `unverified`, which does not fail the command. Absence of evidence isn't evidence of absence.

2. **The rest is measured, not guessed.** `fullPage` still cannot reach inside a region that scrolls on its own — an `overflow: auto` container clips its content, so a row below an inner fold is genuinely absent from the picture and perfectly healthy. `collectVisibilityReport` (`src/utils/visibility.ts`) runs one `page.evaluate` and returns those regions with how much they hide, plus elements `checkVisibility()` calls displayed that nobody can perceive — covered (`elementFromPoint` names the cover), drawn in the colour of their own backdrop, zero-size, or parked outside the document. Both lists go to the judge *and* into the envelope under `### Page visibility`.

Verified against a real Chromium page in `tests/integration/visibility-probe.test.ts` rather than jsdom, where rects and hit-testing don't exist.

src/ai/pilot.ts
DavertMik: Same thread as the other one — addressed in 7dc656e.

Short version: screenshots are already `fullPage: true` (`src/action.ts:107`), so the viewport isn't the limit. But the prompt was treating "absent from the picture" as a contradiction, which produces exactly the false conflict you describe. A conflict now requires the picture to show something *incompatible*; not finding something is `unverified`.

For the case full-page genuinely can't cover — content inside a region that scrolls on its own — `src/utils/visibility.ts` enumerates those regions and how much each hides, and the judge is told never to call a conflict on something inside one.

PR #126, opened 2026-08-19
Give the Tester recorded recipes, not a table of contents

DavertMik:
## The problem

Recorded experience reached the Tester as a **table of contents** — a list of section titles the model could open on demand with `learnExperience`. It almost never did.

From a real 50-test session (Langfuse, 2026-08-19):

| | |
|---|---|
| `<experience>` block (the TOC) present | ~1,300 generations, **49 of 50 tests** |
| `<applied_experience>` (a recipe actually opened) | **1 of 50 tests** |
| `learnExperience` calls | 12 across 50 tests — 8 of them inside just 4 tests |

The cost: nothing one test solved reached the next. The same controls were rediscovered from scratch all run long — one dropdown component's locators were attempted **51 times across 10 different tests**, one menu **32 times across 7** — each attempt paying the same ~60% click failure rate.

## The fix — reuse, not new machinery

The Navigator already solved this. `getSuccessfulExperience()` returns full recipe **bodies**, and `buildResolutionPrompt` wrapped them in an `<experience>` block inline.

1. That inline block moves into `ExperienceTracker.renderExperienceFor(state)`, next to the existing `renderExperienceTocFor`.
2. `navigator.ts` calls the new method. Its prompt is **byte-identical** — the string, the `tag('operation')` log line, and the `isInsideIframe` guard at the call site are all unchanged.
3. `TaskAgent.getExperience()` prefers bodies, falling back to the TOC:

```ts
protected getExperience(actionResult: ActionResult): string {
  const tracker = this.getExperienceTracker();
  return tracker.renderExperienceFor(actionResult) || tracker.renderExperienceTocFor(actionResult);
}
```

`learnExperience` and the TOC path both stay — the TOC is the fallback for pages whose recorded entries hold no reusable recipe.

## Blast radius (verified by grep, not assumed)

`getExperience()` has **exactly three callers**, all heavy-context agents where bodies are appropriate:

- `src/ai/tester.ts:797`
- `src/ai/captain.ts:157`
- `src/ai/driller.ts:373`

Deliberately **untouched**:

- **`src/ai/pilot.ts:702`** — the Pilot does not use `getExperience()`; it calls `renderExperienceTocFor` directly, so its deliberately light conversation keeps its TOC.
- **`src/ai/navigator.ts:699`** — the verification path stays on the TOC.
- **`boat/prima/src/prima.ts:873`** — optional-chained direct TOC call, unaffected.

Also confirmed: `renderExperienceToc` already emits the `<experience>` wrapper and all three callers interpolate the result raw, so there is no double-wrapping.

## Trade-off

The TOC existed to bound token cost, and bodies cost more. `getSuccessfulExperience` already accepts `stripCode` and `includeDescendants` if that scope needs tightening later; no new options were added here.

## Tests

Four unit mocks replace the whole tracker object and so needed `renderExperienceFor` added — two Tester mocks and, less obviously, two Navigator mocks that only stubbed `getSuccessfulExperience`.

- `bun run format` — clean
- `bun run lint:fix` — clean, no fixes
- `bun test tests/unit/` — **1022 pass, 0 fail**
- `bun test tests/integration/` — **80 pass, 1 skip, 0 fail**

This changes what agents see in their prompts, so it may warrant regression coverage — leaving that call to you; I have not applied the label.

### DIFF
--- a/src/ai/tools.ts
+++ b/src/ai/tools.ts
@@ -24,6 +24,7 @@ interface AgentToolDeps extends ToolDeps {
   navigator: Navigator;
   supervisor?: boolean;
   withExperience?: boolean;
+  appliedExperience?: () => string;
 }
 
 export const ASSERTION_TOOLS = ['verify'] as const;
@@ -569,7 +570,7 @@ export function createLearnExperienceTool({ getExperienceTracker, getState }: { 
   });
 }
 
-export function createAgentTools({ explorer, stateManager, ai, researcher, navigator, supervisor, withExperience }: AgentToolDeps): any {
+export function createAgentTools({ explorer, stateManager, ai, researcher, navigator, supervisor, withExperience, appliedExperience }: AgentToolDeps): any {
   const tools: Record<string, any> = {
     see: tool({
       description: dedent`
@@ -799,7 +800,7 @@ export function createAgentTools({ explorer, stateManager, ai, researcher, navig
 
           const previousState = ActionResult.fromState(currentState);
           const actionResult = ActionResult.fromState(currentState);
-          const success = await navigator.resolveState(instruction, actionResult);
+          const success = await navigator.resolveState(instruction, actionResult, { experience: appliedExperience?.() });
 
           const toolResult = await ActionResult.fromState(stateManager.getCurrentState()!).toToolResult(previousState, instruction);
 
--- a/src/ai/tools.ts
+++ b/src/ai/tools.ts
@@ -798,7 +798,7 @@ export function createAgentTools({ explorer, stateManager, ai, researcher, navig
 
           const previousState = ActionResult.fromState(currentState);
           const actionResult = ActionResult.fromState(currentState);
-          const experience = renderExperienceRecipes(explorer.activeTest?.appliedExperience ?? []);
+          const experience = renderExperienceRecipes(explorer.activeTest?.getAppliedExperience(actionResult) ?? []);
           const success = await navigator.resolveState(instruction, actionResult, { experience });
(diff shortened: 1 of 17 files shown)


DenysKuchma (approved):

DenysKuchma (changes requested):
The previous experience index stays in the Pilot conversation when the new page has no recipes. We only clean the tag when tocBlock is truthy, so the Pilot can still see entries from the previous page. Can we always clean the old tag before checking the new block?

src/ai/tools.ts
```diff
@@ -798,7 +798,8 @@ export function createAgentTools({ explorer, stateManager, ai, researcher, navig
 
           const previousState = ActionResult.fromState(currentState);
           const actionResult = ActionResult.fromState(currentState);
-          const success = await navigator.resolveState(instruction, actionResult);
+          const experience = renderExperienceRecipes(explorer.activeTest?.appliedExperience ?? []);
```
DenysKuchma: appliedExperience is test-wide, so this also sends recipes loaded on previous pages and presents them as relevant to the current page. Can we keep the source URL and only pass recipes matching the current state?

DenysKuchma (changes requested):

DenysKuchma:
> The previous experience index stays in the Pilot conversation when the new page has no recipes. We only clean the tag when tocBlock is truthy, so the Pilot can still see entries from the previous page. Can we always clean the old tag before checking the new block?


Been resolved

DenysKuchma (approved):

PR #132, opened 2026-08-21
Compact the error a failed command reports, instead of Playwright's whole retry log

DavertMik:
A failed browser command handed the model Playwright's entire call log — ~850 characters for one disabled Save button: every wait and retry line, `element is not enabled` once per attempt, a class attribute broken across six lines, and ANSI color codes around all of it.

**Before**

```
TimeoutError: click: Timeout 3000ms exceeded.
Call log:
  - waiting for getByRole('button', { name: 'Save' }).first()
  - locator resolved to <button disabled type="button" class="secondary-btn
      btn-text-and-icon

      btn-lg

      disabled">…</button>
  - attempting click action
    2 × waiting for element to be visible, enabled and stable
      - element is not enabled
    - retrying click action
    - waiting 20ms
    … 8 more lines
```

**After** (192 chars)

```
TimeoutError: click: Timeout 3000ms exceeded. - locator resolved to <button disabled type="button" class="secondary-btn btn-text-and-icon btn-lg disabled">…</button> - element is not enabled
```

## The implementation

```js
export function compactErrorMessage(error: unknown): string {
  let text = stripAnsi(String(error));
  for (const strip of STRIP_STRATEGIES) {
    text = strip(text);
  }
  return truncate(text, MAX_COMPACT_ERROR);
}
```

In `src/utils/strings.ts`, next to the `normalizeInlineText` and `truncate` it reuses. It takes anything and stringifies inside, so no call site wraps its error. No regex, no new file, no new module.

`stripCallLog` is the only strategy today: split on the marker, drop `waiting` / `retrying` / `attempting` lines, dedupe the rest with a `Set`, join. Another error shape — a network failure, a stack trace — is a second function in `STRIP_STRATEGIES`, and both the marker and its noise list stay local to the strategy that uses them.

The locator line goes with the other `waiting` lines because the tool result already carries `command` and `locator`; repeating it costs tokens and tells the model nothing new. Errors with no `Call log:` section (`MultipleElementsFound` and its element listing) come back unchanged.

The seven raw `lastError.toString()` / `String(lastError)` sites in click, hover, pressKey, form, back and reset now route through `errorText()`, so that stringification exists in one place.

`clickFailureSuggestion()` reads this text to tell the model whether the element was disabled, covered, hidden or absent. All four markers sit on lines the filter keeps.

## Natural next caller

`navigator.ts:476` trims a failed attempt inline — first non-`at ` line, capped at 220. That is a stack-strip strategy waiting to move into this list, but it changes navigator's output, so it is not in this PR.

## Tests

`bun test tests/integration/` 81 pass. Unit suite passes except the 12 `renderCall` failures from the `asLocator` lookup on `main`, which #131 fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## 2026-08-21
+
+### Changes
+
+- A browser command that fails now reports a short error instead of Playwright's full retry log.
+  Click, hover, key press, form and back/reset failures used to carry every wait and retry line,
+  the same reason repeated once per attempt, and terminal color codes — around 850 characters for
+  one disabled button. What is left is the element the locator resolved to and the reason it could
+  not be acted on, so the AI reads the blocker instead of scrolling past bookkeeping.
+
 ## 2026-08-18
 
 ### Changes
--- a/src/ai/tester.ts
+++ b/src/ai/tester.ts
@@ -15,6 +15,7 @@ import { compactAriaSnapshot, detectFocusArea } from '../utils/aria.ts';
 import { ErrorPageError, isErrorPage } from '../utils/error-page.ts';
 import { createDebug, tag } from '../utils/logger.ts';
 import { loop } from '../utils/loop.ts';
+import { compactErrorMessage } from '../utils/strings.ts';
 import type { Agent, AgentDeps } from './agent.ts';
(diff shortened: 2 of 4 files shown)


DenysKuchma (approved):
Fix tests

PR #133, opened 2026-08-21
Improve tool recovery, API feedback, and cache provenance

DenysKuchma:
## Summary

  - repair tool calls when reasoning models append internal channel markers to valid tool names
  - prevent rejected captured API requests from being presented as valid Fisherman examples
  - classify HTTP failures so Fisherman can choose an appropriate recovery strategy
  - store source URLs with research fingerprints and filter incompatible page families before similarity matching
  - apply the same URL-aware matching when Planner reuses state hashes
  - preserve compatibility with legacy fingerprint files
  - correctly resolve absolute output, knowledge, and experience directories on Windows
  - centralize page-family matching and remove duplicated Planner logic

### DIFF
--- a/src/ai/researcher.ts
+++ b/src/ai/researcher.ts
@@ -139,10 +139,10 @@ export class Researcher extends ResearcherBase implements Agent {
       const combinedHtml = await this.actionResult!.combinedHtml();
 
       if (!deep && !force) {
-        const similar = await findSimilarResearch(combinedHtml);
+        const similar = await findSimilarResearch(combinedHtml, state.url);
         if (similar) {
           tag('operation').log('Similar research found, reusing cached result');
-          if (stateHash) saveResearch(stateHash, similar, combinedHtml);
+          if (stateHash) saveResearch(stateHash, similar, combinedHtml, state.url);
           tag('multiline').log(formatResearchSummary(similar));
           tag('success').log('Research complete (reused)');
           await this.hooksRunner.runAfterHook('researcher', state.url);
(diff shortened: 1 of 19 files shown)


src/utils/url-matcher.ts
```diff
@@ -28,6 +28,21 @@ export function hasDynamicUrlSegment(url: string): boolean {
   return url.split('/').some((seg) => seg.length > 0 && isDynamicSegment(seg));
 }
 
+export function isSamePageFamily(urlA: string, urlB: string): boolean {
+  const partsA = extractStatePath(urlA).split('?')[0].split('#')[0].toLowerCase().split('/').filter(Boolean);
```
DavertMik: use URL for that!

https://developer.mozilla.org/en-US/docs/Web/API/URL_API

src/ai/researcher.ts
```diff
@@ -286,7 +286,7 @@ export class Researcher extends ResearcherBase implements Agent {
 
       let researchFile: string | null = null;
       if (stateHash) {
-        researchFile = saveResearch(stateHash, result.text, combinedHtml);
+        researchFile = saveResearch(stateHash, result.text, combinedHtml, state.url);
```
DavertMik:
```suggestion
        researchFile = saveResearch(state, result.text, combinedHtml);
```

we can just pass state object and extract URL fro it

DavertMik (commented):
4 params in a method call is a smell

probably we need to have some broader value object thet should be passed 

in our case it is State instead of state.url

DavertMik (approved):

PR #142, opened 2026-08-24
Scaffold a config that resolves models without an install

### LINKED ISSUE #134: Issue: @openrouter/ai-sdk-provider is not installed
**Title:** `init` scaffolds a config with an undeclared dependency; error message isn't actionable

Running `npx explorbot init` in a clean directory generates a config that imports `@openrouter/ai-sdk-provider`. That package isn't installed by `init` and isn't declared anywhere, so the next command fails with:

```
Failed to load configuration: Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@openrouter/ai-sdk-provider' imported from ./explorbot.config.js
```

The README's `npm i explorbot --save` step masks this, but only via npm's hoisting of explorbot's own transitive dependency — under pnpm or bun the generated config would fail regardless.

Two small changes would close the gap:

1. `init` installs (or prints the install command for) the provider package it just wrote an import for.
2. Config-loading catches `ERR_MODULE_NOT_FOUND` and reports the missing package with a suggested `npm install` and a pointer to `docs/providers.md`.

Environment: explorbot 0.2.5, Ubuntu, npm.

DavertMik:
Closes #134. Replaces #141, which fixed this by telling the user to run `npm i @openrouter/ai-sdk-provider` — unnecessary, since we already ship that package.

### What happened

`npx explorbot init` in an empty directory. npx put explorbot and its dependencies in `~/.npm/_npx/<hash>/node_modules/`; the project directory got nothing. `init` wrote a config whose first line is `import { createOpenRouter } from '@openrouter/ai-sdk-provider'`. The next command called `import('file:///their-dir/explorbot.config.js')`, and Node resolved that bare specifier **from the file containing it** — their config, in their directory:

```
/their-dir/node_modules/@openrouter/ai-sdk-provider   ✗ empty dir
/their/node_modules/…                                 ✗
/node_modules/…                                       ✗
```

The npx cache is never on that chain, and neither is a `-g` install. Our copy was on their disk the whole time and unreachable from their file. `ERR_MODULE_NOT_FOUND`.

`npm i explorbot --save` hides it: npm flattens the tree and hoists our copy into `./node_modules/` next to explorbot, so the first lookup hits. It works because npm dropped our dependency in their directory, not because we declared it — pnpm and bun keep it nested and fail even with a local install.

### The fix

Stop making their file import anything. We own the config format and we load it, so the config names the model and **we** do the import, from inside our own package:

```javascript
ai: {
  // fast model with tool calling capabilities
  model: 'openrouter/openai/gpt-oss-20b:nitro',
  // vision model for screenshot analysis
  visionModel: 'openrouter/openai/gpt-5.6-luna',
  // agentic model for decision making
  agenticModel: 'openrouter/openai/gpt-5.6-luna',
}
```

`resolveModel()` splits at the first `/`, `PROVIDERS['openrouter'].load()` does `await import('@openrouter/ai-sdk-provider')` in `src/config.ts`. Nothing new — this is the form `init --global` has written since `dcc806a`, with the comment "Models are written as 'provider/model-id' so they resolve without a local node_modules." The global path was already fixed; the local template never got the same treatment. Both now share the same model lines, which also drops a stale `visionModel` the local template had drifted to.

### Still explains a real missing import

A hand-written config that imports a provider package it doesn't have gets a message instead of a stack trace:

```
Configuration imports "@openrouter/ai-sdk-provider", which is not installed.

Install it:
  npm i @openrouter/ai-sdk-provider

Or drop the import and write the model as "provider/model-id" for a provider Explorbot ships: openai, anthropic, google, groq, mistral, openrouter, sambanova

Providers: https://github.com/testomatio/explorbot/blob/main/docs/basics/providers.md
```

Node and bun both report `ERR_MODULE_NOT_FOUND`, and the `.ts` require fallback surfaces it the same way, so one branch in `loadConfig`'s existing catch covers every path. Subpaths are trimmed for the install line (`zod/v4` → `npm i zod`) — a missing module is often transitive, so the message names the specifier that actually failed. Relative specifiers fall through to the generic message; `ConfigMissingError` still short-circuits ahead of it.

### Verified

`init` then `config` in a directory with no `node_modules` and no API key set, resolving all three models. `init --global` unchanged. The missing-package message renders for both `.js` and `.ts` configs. Two unit tests: the scaffolded config has no bare import and loads with resolvable models; a config importing an absent package fails with the package name and install line.

Format, lint, unit (1078) and integration (80) pass. `bun run build` fails locally, identically at `main` — a `node_modules` issue with `playwright-core` needing `chromium-bidi`, unrelated.

### DIFF
--- a/src/commands/init-command.ts
+++ b/src/commands/init-command.ts
@@ -9,12 +9,8 @@ import { log, tag } from '../utils/logger.js';
 import { relativeToCwd } from '../utils/next-steps.ts';
 
 function defaultConfigTemplate(): string {
-  return `// Models are written as 'provider/model-id' so they resolve without installing a provider package.
-// Vercel AI SDK is used to connect to AI providers.
-// Bring your own provider by installing its package and importing it here:
-//   npm i @ai-sdk/openai
-//   import { createOpenAI } from '@ai-sdk/openai';
-//   const openai = createOpenAI({ apiKey: process.env.OPENAI_API_KEY });
+  return `// Models are written as 'provider/model-id' — Explorbot bundles these providers, nothing to install.
+// To use another one, install its Vercel AI SDK package and import it here.
 // https://github.com/testomatio/explorbot/blob/main/docs/basics/providers.md
 
 const config = {
--- a/src/commands/init-command.ts
+++ b/src/commands/init-command.ts
@@ -9,8 +9,7 @@ import { log, tag } from '../utils/logger.js';
 import { relativeToCwd } from '../utils/next-steps.ts';
 
 function defaultConfigTemplate(): string {
-  return `// Models are written as 'provider/model-id' — Explorbot bundles these providers, nothing to install.
-// To use another one, install its Vercel AI SDK package and import it here.
+  return `// 'provider/model-id' uses a bundled provider. Or import a Vercel AI SDK one.
 // https://github.com/testomatio/explorbot/blob/main/docs/basics/providers.md
 
 const config = {
--- a/src/commands/init-command.ts
+++ b/src/commands/init-command.ts
@@ -9,7 +9,8 @@ import { log, tag } from '../utils/logger.js';
 import { relativeToCwd } from '../utils/next-steps.ts';
 
 function defaultConfigTemplate(): string {
-  return `// 'provider/model-id' uses a bundled provider. Or import a Vercel AI SDK one.
+  return `// 'provider/model-id' uses a bundled provider.
+// It is also possible to import provider as a module from Vercel AI SDK.  
 // https://github.com/testomatio/explorbot/blob/main/docs/basics/providers.md
 
 const config = {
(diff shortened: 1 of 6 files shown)


src/commands/init-command.ts
```diff
@@ -8,15 +8,9 @@ import { getCliName } from '../utils/cli-name.ts';
...
-// import { '<your provider here>' } from '<your provider package here>';
-
-// Vercel AI SDK is used to connect to AI providers.
-// Bring your own provider or use OpenRouter (one API key, many providers).
-// https://github.com/testomatio/explorbot/blob/main/docs/providers.md
-const openrouter = createOpenRouter({
-  apiKey: process.env.OPENROUTER_API_KEY,
-});
+function defaultConfigTemplate(): string {
+  return `// 'provider/model-id' uses a bundled provider. Or import a Vercel AI SDK one.
```
DavertMik:
```suggestion
  return `// 'provider/model-id' uses a bundled provider.
// It is also possible to import provider as a module from Vercel AI SDK.  
```

PR #143, opened 2026-08-25
Stabilize AI requests and isolate tests after error pages

DenysKuchma:
## Summary

  Improves AI request reliability, restores vision processing, makes Pilot verdicts more evidence-based, and prevents an error page left by one test from incorrectly failing the following tests.

  ## Changes

  - Limit concurrent AI requests through `ai.maxParallelRequests` to reduce provider rate-limit bursts.
  - Honor `ai.retryAttempts` and `ai.retryDelay` configuration.
  - Rename the obsolete `ai.maxAttempts` AI setting to `ai.retryAttempts`.
  - Send screenshots to vision models in the format expected by the AI SDK.
  - Disable vision for the rest of the session after a vision failure to avoid repeated failing requests.
  - Support model narration through a no-op `commentary` tool.
  - Repair Harmony channel-style tool calls instead of rejecting the whole generation.
  - Exclude narration calls from recorded tool execution evidence.
  - Prevent successful context and research calls from being treated as proof that a test goal was achieved.
  - Clarify Pilot evidence so a successfully executed check can still prove that the scenario failed.
  - Recover from an error page left by the previous test before building context for the next test.
  - Preserve genuine error reporting when the new test’s own start page is an error page.
  - Always generate the session analysis report, including when exploration exits with an error.

### DIFF
--- a/src/ai/pilot.ts
+++ b/src/ai/pilot.ts
@@ -25,7 +25,8 @@ import { capabilityGroundingRule, dataProtectionRules } from './rules.ts';
 import { isInteractive } from './task-agent.ts';
 import { withdrawVisionTools } from './tools.ts';
 
-const CHECK_TOOLS = ['verify', 'see', 'research', 'context'];
+const CHECK_TOOLS = ['verify', 'see', 'research'];
+const EVIDENCE_TOOLS = ['verify', 'see'];
 const META_TOOLS = ['record', 'reset', 'stop', 'finish'];
 const PILOT_MESSAGE_LIMIT = 2;
 const PILOT_MESSAGE_MAX_LENGTH = 160;
@@ -326,6 +327,10 @@ export class Pilot implements Agent {
       overrides the others — weigh them together. Tester's record() notes are the LEAST reliable; always
       cross-check against actual actions and state. Visual screenshot analysis is strong for UI state
       (active tabs, visible counts, colors).
+      Judge every check by WHAT IT ESTABLISHES, never by the fact that it ran. A check that executed
+      successfully is failure evidence when its content negates the scenario goal — the goal's object
+      absent, the action not performed, the interaction impossible. "The check passed" and "the goal was
+      met" are different claims.
       If the final page clearly shows an equivalent success state in a different UI form, do not fail only
       because one narrow assertion targeted a specific badge, count, toast, or wording that the product
       represents differently.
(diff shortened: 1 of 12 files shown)


src/ai/provider.ts
```diff
@@ -154,10 +172,28 @@ export class Provider {
   private getRetryOptions(options: any = {}): RetryOptions {
     return {
       ...this.defaultRetryOptions,
-      maxAttempts: options.maxRetries || this.defaultRetryOptions.maxAttempts,
+      maxAttempts: options.maxRetries || this.config.retryAttempts || this.defaultRetryOptions.maxAttempts,
+      baseDelay: this.config.retryDelay || this.defaultRetryOptions.baseDelay,
     };
   }
 
+  private async withModelSlot<T>(fn: () => Promise<T>): Promise<T> {
```
DavertMik: not really get what it is

src/ai/provider.ts
```diff
@@ -20,6 +22,20 @@ const responseLog = createDebug('explorbot:provider:in');
 class AiError extends Error {}
 export class ContextLengthError extends Error {}
 
+const DEFAULT_PARALLEL_REQUESTS = 4;
+
+const modelSlotContext = new AsyncLocalStorage<boolean>();
+
+const HARMONY_CHANNELS = ['commentary', 'analysis', 'final'];
+
+function createCommentaryTool() {
```
DavertMik: do we need this? 
model reasons a different way

src/ai/pilot.ts
```diff
@@ -574,8 +579,8 @@ export class Pilot implements Agent {
   }
 
   async settleExpectations(task: Test, finalState?: ActionResult): Promise<SettledExpectation[]> {
-    let image: string | null = null;
-    if (finalState?.screenshot && this.provider.hasVision()) image = `data:image/png;base64,${finalState.screenshot.toString('base64')}`;
+    let image: Buffer | null = null;
+    if (finalState?.screenshot && this.provider.hasVision()) image = finalState.screenshot;
```
DavertMik, line 583: is this supported by all Vercel AI SDK providers or openoruter/luna specific ?

DavertMik (changes requested):

src/ai/pilot.ts
DenysKuchma: Yes, this is supported by the Vercel AI SDK rather than being Luna/OpenRouter-specific. A Buffer is a Uint8Array, which is accepted as file data and normalized by the provider adapter. The selected model still needs vision support

src/ai/provider.ts
DenysKuchma: We need compatibility handling for the Harmony/gpt-oss output because these models sometimes emit analysis/commentary as tool calls, causing the AI SDK to reject the whole generation

src/ai/provider.ts
DenysKuchma: It is a concurrency limiter for AI requests. Explorbot can run several agents and tests in parallel, which previously caused request bursts, rate limits, and synchronized retries. withModelSlot allows only ai.maxParallelRequests model operations at once and queues the rest.

PR #146, opened 2026-08-26
Detect unmarked modals at capture so the Tester scopes interactions i…

DenysKuchma:
## Summary

Modals without `role="dialog"`/`aria-modal` were invisible to focus-area detection, so with a dialog open the Tester kept interacting with the page behind it (session `YtterbicRunningTeal168`: `focus_scope=0` all night). Now every state capture probes the live DOM for open overlays using the Driller's existing extractor, and the state carries a resolved `overlay` descriptor that the Tester, Pilot and deep analysis read directly instead of re-deriving it from ARIA.

  Class-name matches only count when the element is truly floating (`fixed`/`absolute`/z-index), so a sticky header named "overlay" can't become a false modal; the expensive full-page geometry scan is not run on every action — it stays in the Driller, which also gains the class tokens and the nested-overlay dedup through the shared extractor. Covered by 8 new unit tests; unit suite at 1093 pass with only the pre-existing flaky `remote` failures.

### DIFF
--- a/src/action-result.ts
+++ b/src/action-result.ts
@@ -2,11 +2,12 @@ import fs from 'node:fs';
 import { join } from 'node:path';
 import { ConfigParser, type HtmlConfig, outputPath } from './config.ts';
 import type { Link, WebPageState } from './state-manager.ts';
-import { type FocusAreaResult, LARGE_ARIA_CHANGE_THRESHOLD, compactAriaSnapshot, detectFocusArea, diffAriaSnapshots } from './utils/aria.ts';
+import { LARGE_ARIA_CHANGE_THRESHOLD, compactAriaSnapshot, diffAriaSnapshots } from './utils/aria.ts';
 import { TTLCache } from './utils/cache.ts';
 import { type HtmlDiffPart, type HtmlDiffResult, htmlDiff, liveRegionMessages } from './utils/html-diff.ts';
 import { extractHeadings, extractLinks, extractTargetedHtml, htmlCombinedSnapshot, htmlMinimalUISnapshot, htmlTextSnapshot, minifyHtml } from './utils/html.ts';
 import { createDebug } from './utils/logger.ts';
+import { Overlay } from './utils/overlay.ts';
 import { slugify } from './utils/strings.ts';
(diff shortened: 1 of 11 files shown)


src/action.ts
```diff
@@ -144,11 +144,31 @@ class Action {
...
+              config: {
+                interactiveContentSelector: HTML_SELECTORS.interactiveContent,
+                limits: HTML_EXTRACTION_LIMITS,
+                overlaySelectors: HTML_SELECTORS.modalOverlays,
+                overlaySemanticSelector: HTML_SELECTORS.overlaySemanticSelector,
+                visibilityLimits: HTML_VISIBILITY_LIMITS,
+                geometryFallback: false,
+              },
+            }
+          );
```
DavertMik: can we somehow simplify this part or move it to like private method?
or dedicated function
this config thing is nosiy

src/ai/tester.ts
```diff
@@ -531,7 +530,7 @@ export class Tester extends TaskAgent implements Agent {
 
     let context = '';
 
-    const focusArea = detectFocusArea(currentState.ariaSnapshot);
```
DavertMik, line 534: we may still need focus area 
focus area is where the input is or where input selected

src/action-result.ts
```diff
@@ -131,6 +133,14 @@ export class ActionResult implements ActionResultData {
       this._ariaSnapshot = data.ariaSnapshot;
     }
 
+    if (data.overlayHtml) {
+      const headings = extractHeadings(data.overlayHtml);
+      const name = [headings.h1, headings.h2, headings.h3, headings.h4].filter(Boolean).join(' ');
+      this.overlay = { detected: true, type: 'modal', name: name || null };
```
DavertMik: seems like we have need for Overlay class

```js
this.overlay = new Overlay({ name: })
```
like this one

DavertMik (changes requested):
It seems like we need to create a new utils/overlay class and move all data logic into that one

We night not need that full match url logic soon for web apps that do correct `role=dialog` mapping so I think there is a chance we may need to disable this at one point

DavertMik (approved):

PR #158, opened 2026-08-29
Detect modals, drawers and soft navigation as region-of-interest states

DavertMik:
## What

After every action, the page diff is checked for a newly appeared subtree ≥ 10K chars of minified HTML. A Playwright geometry probe (`elementFromPoint` sampling, viewport coverage, inert siblings, scroll lock) then classifies it: **overlaying** (modal/drawer — page behind not actionable) or **inline region** (soft navigation — area of interest, page still usable). Detection is fully structural — no AI in the path, no class-name selector heuristics.

## Key changes

- **`overlay.ts` is now the single overlay-detection module**: `findAppearedSubRoot`, `probeRegionCoverage` + `classifyRegionCoverage`, `Overlay.fromSubRoot`. `Overlay` gains `root`, `present`, and `drawer`/`region` types; `detected` keeps meaning "verified overlaying".
- **State identity**: named regions enter the state hash (`region_<name>`), so modal open *and* close record transitions — visible to dead-loop detection. New `baseHash` keeps the research cache keyed to the page.
- **Tool results**: instead of `...collapsed (12000 chars)...`, an opened region reports `drawer "Edit User" opened, scope: aside.detail-panel` plus the region's snapshot (`pageDiff.areaOfInterest`).
- **Tester**: `<focus_scope>` carries the concrete root selector for verified overlays; inline regions get a softer `<area_of_interest>` block that leaves page navigation actionable.
- **Pilot**: `<state>` distinguishes `modal:` (overlaying, with root) from `region:` (inline).
- **Experience envelope**: region-state files carry `root: <selector>` frontmatter; records with `root` load only while a matching region is open.
- **Removed**: the selector-heuristic path — `extractVisibleOverlayHtml`, `OVERLAY_SELECTORS`, `Action.captureOverlayHtml`, the `overlayHtml` chain, and Driller's private extractor (Driller now reads nested overlays from `pageDiff.htmlParts`).

Trade-off (accepted): an overlay already open at the very first capture with no ARIA role isn't detected until the next action — in exchange for deleting class-name pattern matching that only worked on sites that name their CSS a certain way.

Spec: `docs/superpowers/specs/2026-08-29-region-states-design.md` · Plan: `docs/superpowers/plans/2026-08-29-region-states.md`

## Tests

- 1137 unit + 83 integration tests pass locally.
- New coverage: appeared-subroot detection, coverage classifier, region hash/baseHash, diff memoization + `areaOfInterest`, StateManager region transitions, experience `root` write/gating, Pilot state context.

https://claude.ai/code/session_01TkmQDF4DkphEYoy71twQNx

### DIFF
--- a/src/action-result.ts
+++ b/src/action-result.ts
@@ -262,6 +262,7 @@ export class ActionResult implements ActionResultData {
 
   isRelevantExperienceRecord(record: WebPageState, options?: { includeDescendantExperience?: boolean }): boolean {
     if (!record.url || !this.url) return false;
+    if (record.region && this.overlay.name !== record.region) return false;
     if (record.root) {
       if (!this.overlay.present) return false;
       if (this.overlay.root && this.overlay.root !== record.root) return false;
--- a/src/action-result.ts
+++ b/src/action-result.ts
@@ -554,8 +554,8 @@ export class ActionResult implements ActionResultData {
       if (subtree.length > HTML_PART_SUBTREE_BUDGET) {
         subtree = `${subtree.slice(0, HTML_PART_SUBTREE_BUDGET)}...<!-- truncated -->`;
       }
-      pageDiff.htmlParts = [{ container: this.overlay.root, subtree, added: [], removed: [] }];
-    } else if (diff.htmlParts.length > 0) {
+      pageDiff.htmlParts = [{ container: this.overlay.root, subtree, rawSize: subtree.length, added: [], removed: [] }];
+    } else if (diff.isSameUrl() && diff.htmlParts.length > 0) {
       const collapsed = collapseHtmlParts(await diff.cleanedHtmlParts());
       if (collapsed.length > 0) {
         pageDiff.htmlParts = collapsed;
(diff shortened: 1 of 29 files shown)


src/action.ts
```diff
@@ -195,14 +193,62 @@ class Action {
...
+      if (await overlayPage.isStillOpen(previousOverlay)) {
+        carried = previousOverlay;
+      } else {
+        debugLog(`Region closed: ${previousOverlay.name || previousOverlay.type}`);
+        const parent = previousOverlay.parent;
+        if (parent?.xpath) {
+          const restored = new Overlay(parent);
+          if (await overlayPage.isStillOpen(restored)) result.overlay = restored;
+        }
+        return;
```
DenysKuchma: If one dialog closes and another opens in the same action, this returns before detecting the replacement

src/action-result.ts
```diff
@@ -260,6 +262,10 @@ export class ActionResult implements ActionResultData {
 
   isRelevantExperienceRecord(record: WebPageState, options?: { includeDescendantExperience?: boolean }): boolean {
     if (!record.url || !this.url) return false;
+    if (record.root) {
+      if (!this.overlay.present) return false;
+      if (this.overlay.root && this.overlay.root !== record.root) return false;
```
DenysKuchma, line 268: If the current region has no root, this accepts every root-scoped experience for the same page. Should a record with root require an exact current root match?

DenysKuchma (changes requested):

PR #159, opened 2026-08-29
Skip planning on error and not-found pages

DavertMik:
Exploration burned a full planning round on pages with nothing to test — including the case in the reported trace, a 200 response whose only sign of being dead is its own copy.

## Changes

**`explore-command.ts` — sub-page skip.** `discoverNewSubPages` checks the page right after visiting it. If `getStateErrorPageError` reports an error page: warn, mark it failed, move to the next candidate. No research, no planning, no AI call spent on it. This covers server-rendered errors — anything with a 4xx/5xx document status or a standard HTTP error title.

**`planner.ts` — one prompt line.** That check only sees HTTP status and standard error titles, so a page returning 200 with app-specific "doesn't exist" copy slips past it. Reading that from arbitrary wording is AI judgment, so the planning prompt carries the rule (general, no wording from any specific page): a page that reports a missing resource, shows a failure state, or holds no content and no controls gets an empty scenarios list, and nothing is invented for it.

**Empty output is a skip, not a failure.** Zero scenarios with no plan to expand used to throw the same generic planning error as a failed call, so exploration retried the page and then spent two more calls per remaining planning style on it — the prompt rule above bought nothing. An empty scenario list is a value now, not an exception: the planner returns the plan it built, and `runAllStyles` reads its length. No tests means nothing to test here, so it stops trying further styles and the next sub-page candidate is picked; the empty plan is not carried into the saved plans. An expanding plan is untouched — a later style returning nothing keeps the tests already planned.

## Tests

`tests/integration/planner.test.ts` asserts the prompt carries the rule, that empty output yields an empty plan rather than an error, and that an existing plan survives an empty style.

`bun test tests/integration/` and `bun test tests/unit/` pass, format and lint clean.

The prompt half can only be confirmed in a live run — worth the `regression` label if you want it proven against the page in the trace.

https://claude.ai/code/session_013TYaLWuuLjj5mb26NuYjoj

### DIFF
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -195,10 +195,6 @@ export class Planner extends PlannerBase implements Agent {
         throw new Error('No tasks were created successfully');
       }
 
-      if (aiResult.object.scenarios.length === 0 && !this.currentPlan) {
-        throw new EmptyPlanError(actionResult.url || state.url);
-      }
-
       const defaultStartUrl = this.getDefaultStartUrl(state);
       const fromPlanning = aiResult.object.scenarios.map((s: any) => new Test(s.scenario, s.priority, s.expectedOutcomes, s.startUrl || defaultStartUrl, s.steps || []));
 
@@ -632,12 +628,5 @@ export class Planner extends PlannerBase implements Agent {
       8. For a fresh plan propose at least ${this.MIN_TASKS} tests; when expanding an existing plan return ONLY new scenarios not in the tested list, and an empty array if none remain.
     </task>
     `;
-  }
-}
-
-export class EmptyPlanError extends Error {
-  constructor(public readonly url: string) {
-    super(`no scenarios proposed for ${url}`);
-    this.name = 'EmptyPlanError';
   }
 }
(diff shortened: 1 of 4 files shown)


src/ai/planner.ts
```diff
@@ -335,6 +335,7 @@ export class Planner extends PlannerBase implements Agent {
       <task>
       Based on the page research, create ${this.MIN_TASKS}-${this.MAX_TASKS} exploratory testing scenarios.
       For each scenario provide specific steps and expected outcomes.
+      Exception: if the page reports the requested resource is missing, shows a failure state, or holds no content and no controls, return an empty scenarios list. Never invent tests for a page with nothing to exercise.
```
DenysKuchma, line 334: plan() still throws when scenarios is empty and there is no current plan. So this turns the intended skip into a planning error. Should an empty list be handled as a valid skip?

DenysKuchma (changes requested):

src/ai/planner.ts
DavertMik: Reworked in 64ad7e7 — the earlier `EmptyPlanError` is gone. An empty scenario list is a value, not an exception: `plan()` returns the plan it built and `runAllStyles` checks `plan.tests.length === 0`, which means nothing to test here, so it stops trying further planning styles and the loop picks the next sub-page candidate. An empty plan is not pushed into the saved plans either.

Expanding an existing plan is unchanged, and a genuine planning failure still retries as before.

DenysKuchma (approved):

PR #160, opened 2026-08-29
Make Fisherman's replicate mode trustworthy

DavertMik:
Fixes the five root causes behind every failed Fisherman episode in the April/June traces: the agent poisoning its own request store, first-match spec lookups, a scope filter that never matched, an unconditionally successful `finish`, and silent exhaustion.

Spec: `docs/superpowers/specs/2026-08-29-fisherman-reliability-design.md` · Plan: `docs/superpowers/plans/2026-08-29-fisherman-reliability.md`

## What changed

- **Replicate mode no longer eats its own rejects.** `loadFromDisk()` admits only `xhr_*` files — provenance was already encoded in the id prefix (browser captures `xhr_`, Fisherman's own calls unprefixed), so no new envelope key and poisoned directories migrate for free. **Behavioral change:** a `output/requests/` dir containing only Fisherman-made files now reports replicate mode unavailable — correct, there was never real browser traffic to replicate.
- **Ranked example lookup.** `findCapturedRequest` prefers exact endpoint over deeper sub-path, then 2xx over rejection, then newest. Paths are normalized with the existing `isDynamicSegment()`, so `{id}` patterns and concrete ids both match.
- **Scope that actually matches.** The endpoint list is scoped by the most selective non-id path segment shared between the page URL and captured request paths (the project slug wins over generic literals). When scoping degrades to the global list, the prompt says so instead of silently listing every project's endpoints.
- **`finish` gated by the request ledger.** Success requires at least one real 2xx write in this run; claimed created ids are verified against the requests that actually returned them (`RequestMap` in `src/utils/request-map.ts`), unverifiable claims are dropped, and each verified item names its creating request (`request: "POST /api/…"`, logged by Pilot as `via POST /api/…`). A run that ends without `finish` synthesizes an honest summary from the ledger instead of returning `summary: ''`.
- **Bounded guessing.** Four consecutive failures against one endpoint end the run; one general prompt rule forbids creating a substitute resource type when the requested one has no endpoint.
- New integration test (`tests/integration/fisherman.test.ts`, aimock) covers the scoped prompt and the ledger gate end-to-end; unit coverage for ranking, scoping, load filtering, and ledger results.

## Live-session auth (2026-08-30)

Langfuse trace `de95bd1cffce09169599d99d1bee56cd` (beta.testomat.io) showed every Fisherman write rejected `403 {"error":"Unauthorized"}` in ~8ms while the same browser session performed successful writes to the same project minutes earlier. The requests were doomed by Fisherman's own credential assembly: an unfiltered cookie-jar dump with a dead localhost session pair ahead of the valid one, plus an `x-csrf-token` scraped from a 7-week-old capture of another project (the store's "newest" iteration was really alphabetical filename order).

Principle now enforced (spec: `docs/superpowers/specs/2026-08-30-fisherman-live-session-auth-design.md`): **captures are a source of API shape, never credentials.** Credentials come from the live browser session or explicit `api.headers` config.

- `extractAuthHeaders` is session-gated: only captures made during this session, newest first by timestamp; `cookie` is never scraped from captures; capture files without timestamps read as stale.
- `refreshAuth` precedence is captured < live browser < config, and the browser-derived layers apply only in **replicate mode** — achieve mode authenticates solely through `api.headers`. **Behavioral change:** `Authorization`/`x-api-key` values from previous sessions' captures are no longer sent; long-lived tokens belong in `api.headers`.
- The cookie jar is filtered to the API origin via `context.cookies(baseEndpoint)`; the live CSRF token is read from the page's `meta[name="csrf-token"]` (provider renamed to `browserHeaderProvider`). A failed CSRF read degrades to no token without discarding the cookies.
- `isDynamicSegment` gained a version-segment carve-out: `v1`/`v2`-style segments are never treated as ids, so `POST /api/v1/suites` stays literal in endpoint lists.
- A turn without a tool call is now treated as `finish` (tool choice no longer forced): gpt-oss-style models that close with prose or a `json` pseudo-tool no longer 400 the whole invocation — which previously discarded the completed write roundtrips and re-created the same data on every retry (4 duplicate suite+test pairs in trace `7d5ddba053a054f6a3725503c9569340`).

Review follow-ups deliberately left out of this PR: dedupe of the run-ledger watermark (`fisherman.ts` / `fisherman-tools.ts`), `RequestMap`'s last-write-wins attribution when two responses return the same id (report text only), redacting auth headers in the persisted `output/requests/*.yaml` files (pre-existing behavior), and `oauth2`/hex-word segments still classifying as dynamic.

## Regression fixture: deliberately NOT enabled

The plan's rollout step called for flipping `fisherman: { enabled: true }` in `tests/regression/fixture/explorbot.config.js`, gated on verifying the fixture can feed replicate mode. It cannot: the Trackly fixture has **zero fetch/XHR** — all UI writes are native form posts (`form.requestSubmit()` in `client.js`), which `XhrCapture` ignores by resource type and content type. Replicate mode would see zero captures, so the flip would prove nothing. Options:

- **(a)** Add an `api` block to the regression config — the fixture already serves `/api/openapi.json`, so achieve mode works today, but that exercises a different code path than these fixes target.
- **(b)** Change the fixture client to submit forms via `fetch` with JSON — exercises replicate mode, but alters fixture behavior for every regression scenario.
- **(c)** Leave Fisherman disabled in regression for now.

Until one of those lands, the plan's acceptance criterion (a trace where `precondition()` ids appear on the page the Tester acts on) is blocked on that decision. Regression runs remain user-triggered only — apply the `regression` label when you want one.

https://claude.ai/code/session_01MHhNTXzihYU8f8mqNdjrVP
https://claude.ai/code/session_019JP2duAZr2ppH436viAJYs

### DIFF
--- a/src/ai/fisherman-tools.ts
+++ b/src/ai/fisherman-tools.ts
@@ -18,6 +18,12 @@ export function createFishermanTools(apiClient: ApiClient, requestStore: Request
   const successfulWrites = () => runRequests().filter((r) => r.isWrite && !r.error && r.status >= 200 && r.status < 400);
   const getResult = () => result ?? synthesizeResult(runRequests(), successfulWrites());
   const isFinished = () => finished;
+  const finishFromText = (text?: string) => {
+    finished = true;
+    const synthesized = synthesizeResult(runRequests(), successfulWrites());
+    if (text && synthesized.success) synthesized.summary = text;
+    result = synthesized;
+  };
 
   const tools = {
     getEndpointSpec: tool({
@@ -198,7 +204,7 @@ export function createFishermanTools(apiClient: ApiClient, requestStore: Request
     }),
   };
 
-  return { tools, getResult, isFinished };
+  return { tools, getResult, isFinished, finishFromText };
 }
 
 function synthesizeResult(made: RequestResult[], writes: RequestResult[]): FishermanResult {
--- a/src/ai/fisherman-tools.ts
+++ b/src/ai/fisherman-tools.ts
@@ -6,6 +6,7 @@ import type { RequestResult } from '../api/request-result.ts';
 import type { RequestStore } from '../api/request-store.ts';
 import { extractEndpointDefinition } from '../api/spec-reader.ts';
 import { tag } from '../utils/logger.ts';
+import { RequestMap } from '../utils/request-map.ts';
 import { isDynamicSegment } from '../utils/url-matcher.ts';
 
 export function createFishermanTools(apiClient: ApiClient, requestStore: RequestStore, opts: { spec?: any; baseEndpoint?: string }) {
@@ -159,12 +160,7 @@ export function createFishermanTools(apiClient: ApiClient, requestStore: Request
           return { finished: false, error: 'No successful write request was made in this run, so nothing was created. Keep working, or call stop if the data cannot be prepared.' };
         }
 
-        const viaById = new Map<string, string>();
-        for (const write of writes) {
-          const { id } = write.extractIdAndTitle();
-          if (id === undefined) continue;
-          viaById.set(String(id), `${write.method} ${write.path}`);
-        }
+        const createdRequests = new RequestMap(writes);
 
         const verified: FishermanResult['created'] = [];
         for (const item of created) {
@@ -172,12 +168,12 @@ export function createFishermanTools(apiClient: ApiClient, requestStore: Request
             verified.push(item);
             continue;
           }
-          const via = viaById.get(String(item.id));
-          if (!via) {
+          const request = createdRequests.get(item.id);
+          if (!request) {
             tag('warning').log(`Fisherman: dropped unverified created item ${item.type} (id: ${item.id})`);
             continue;
           }
-          verified.push({ ...item, via });
+          verified.push({ ...item, request: request.toEndpoint() });
         }
         if (verified.length === 0) verified.push(...writes.map(toCreatedItem));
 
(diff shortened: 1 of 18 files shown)


src/ai/fisherman-tools.ts
```diff
@@ -169,7 +204,21 @@ export function createFishermanTools(apiClient: ApiClient, requestStore: Request
...
-  return { tools, getResult, isFinished };
+  return { tools, getResult, isFinished, finishFromText };
+}
+
+function synthesizeResult(made: RequestResult[], writes: RequestResult[]): FishermanResult {
+  const failures = made.filter((r) => r.status >= 400 || r.error);
+  let summary = `Stopped before finishing: ${made.length} requests, ${writes.length} successful writes, ${failures.length} failed`;
+  const lastFailure = failures[failures.length - 1];
+  if (lastFailure) summary += `; last failure: ${lastFailure.toSummary()}`;
+  return { success: writes.length > 0, summary, created: writes.map(toCreatedItem), failed: [] };
```
DenysKuchma: If the run stopped before finish, shouldn’t this stay false? Otherwise one successful write makes a partially completed precondition look ready to Pilot

src/api/request-store.ts
```diff
@@ -148,5 +186,8 @@ export class RequestStore {
 }
 
 function normalizePathPattern(urlPath: string): string {
-  return urlPath.replace(/\/[0-9a-f]{24}\b/g, '/{id}').replace(/\/\d+\b/g, '/{id}');
+  return urlPath
+    .split('/')
+    .map((segment) => (segment && isDynamicSegment(segment) ? '{id}' : segment))
```
DenysKuchma, line 197: This also treats words like feed, cafe, or dead as IDs, so /api/feed becomes /api/{id}. We probably need stricter ID detection for API paths

src/api/request-store.ts
```diff
@@ -137,7 +160,22 @@ export class RequestStore {
...
+    const writes = this.capturedRequests.filter((r) => writeMethods.has(r.method));
+    const scopeSegments = scopePath.split('/').filter(Boolean);
+    if (scopeSegments.length === 0) return writes;
+
+    let scopeKey = '';
+    let fewest = Number.POSITIVE_INFINITY;
+    for (const segment of scopeSegments) {
+      if (isDynamicSegment(segment)) continue;
+      const matches = writes.filter((r) => r.path.split('/').includes(segment)).length;
+      if (matches === 0 || matches >= fewest) continue;
```
DenysKuchma: On a tie this keeps the first segment, which can be something generic like projects instead of the actual project slug. That can scope requests to the wrong project

DenysKuchma (changes requested):

PR #165, opened 2026-08-30
Let the plan name the site an explorbot test run explores

DavertMik:
Running a saved plan against a global installation was impossible: the command died in config loading before it ever opened the plan.

```
$ explorbot test ~/plans/pr-3625-run-user-rendering.md
Configuration loaded from: /home/davert/.explorbot/config.js
Failed to start: Failed to load configuration: Error: No site to explore. Pass a URL to the command or set EXPLORBOT_URL.
```

`~/.explorbot/config.js` holds models and keys, never a site — every command names the site it runs against. `test` named none: it built its `ExplorBot` with `from: undefined`, so `enterGlobalMode` had no target. The plan carries the URL all along, in its `### Prerequisite` section, so the CLI now reads it before the container is built and passes it as the run target. The same target feeds the `EXPLORBOT_*` branch of config loading, so a config-free env setup gets the fix too.

`src/commands/test-command.ts` is unchanged — the gap was in bootstrapping, not in test selection.

Also here:

- `explorbot test <planfile>` with no index runs every enabled test, as `--help` and the docs already said. It used to run only the first pending one.
- A plan's URL — its own, else the first test's — becomes `Plan.startUrl`, replacing the four copies of the expression (CLI twice, both historian writers).

### Verification

Smoke run against a real plan with an out-of-range index, so nothing executed past navigation:

```
Global mode: https://beta.testomat.io stored in ~/.explorbot/sites/beta.testomat.io
Plan loaded: "Run assignee rendering and actions" (8 tests, 8 pending)
Navigating to https://beta.testomat.io/projects/.../runs
Navigation resolved successfully
Failed: Test #99 not found. Available: 1-8
```

`bun test` on test-plan, historian, config and global-config: 97 pass.

### Notes

- One error-path change: a plan with no Prerequisite URL whose tests are all finished now navigates first and reports "All tests are already complete" instead of "No URL found in plan or tests".
- `rerun <file>` has the same gap and is left alone here.

https://claude.ai/code/session_013W3pakMDZuDEjzm6LCFMCB


bin/explorbot-cli.ts
```diff
@@ -325,29 +325,32 @@ addCommonOptions(program.command('plan:load <planfile> [index]').description('Lo
...
-      await explorBot.start();
-
       let planfileArg = planfile;
       let indexArg = index;
       if (options.fromPlan) {
         planfileArg = options.fromPlan;
         indexArg = planfile;
       }
 
+      const planFile = [planfileArg, `${planfileArg}.md`].find((file) => fs.existsSync(file));
```
DenysKuchma: This only checks the current directory, while loadPlan() also resolves names through getPlansDir(). In global mode, explorbot test saved-plan can therefore fail before loadPlan() even though the plan exists. Could both paths use the same resolver?

DenysKuchma (changes requested):

bin/explorbot-cli.ts
DavertMik: Both paths go through one resolver now — `resolvePlanPath()` in `src/utils/plan-path.ts` (7427b7c). It takes a path or a bare name, `.md` optional, and looks in the working directory first, then in a plans directory.

The pre-flight peek runs before the config is loaded, so it has no plans directory to name; it searches the plans directory of every registered site instead (`listSites()`), which is enough to read the plan, take its URL and let config loading resolve the site from it. `ExplorBot.resolvePlanPath()` keeps its signature and delegates with `this.getPlansDir()`, so once the site is known only that site's plans are searched.

Verified against a temp `$HOME`: a global installation with no `EXPLORBOT_URL` and `saved-plan.md` under `~/.explorbot/sites/app.example.com/output/plans/` — `explorbot test saved-plan` used to die with "No site to explore", now resolves the plan, enters global mode on `https://app.example.com`, and `loadPlan()` resolves to the same file. Unit tests in `tests/unit/plan-path.test.ts`.

`plan:load` and `rerun` keep their own cwd-only resolution — same gap, left out of this PR like the `rerun` note in the description.

PR #171, opened 2026-09-01
Plan one verified operation per test, and stop calling test data disposable

DavertMik:
Traced from the last planner runs on Langfuse. Supersedes #168 — see below for what was taken from it and what was not.

## What the traces showed

Trace `d5d1f55a` (`planner: /projects/.../`) produced:

> Create a disposable test suite, update its title, and delete it while verifying each lifecycle state.

Two prompt lines multiplied together to force that:

- `planner.ts:121` — *"If two scenarios cannot run independently (one requires the other to run first), merge them into one."*
- `rules.ts` protectionRule — *"the same scenario must first create a disposable target and then destroy that specific target"*

Update and delete each require a create, so the first rule literally instructed merging all three. `rules/planner/styles/normal.md:4` reinforced it with an arrow chain read as one test's steps, and `planner.ts:351` only said *"plan them in correct order: create, read, update"* — never that they were separate tests, and it omitted delete.

**"disposable" did more damage than an ugly title.** Trace `f18afa62` produced *"Edit a disposable existing plan…"* whose first step is **"Open the plans list containing the disposable plan created for this scenario"** — with no step that creates it. The planner read the word as a *kind of record that already exists* rather than as an instruction, so the scenario is unexecutable. (`git log -S disposable` shows the word was only ever added, in 09a9847 and c1950ef — never removed.)

## The fix

The old rule described a *shape* ("never split a workflow"), which gives no way to tell a multistep test from a lifecycle chain. The new rule names a *unit*, and counts **verified** operations rather than performed ones:

```
One test verifies ONE business operation: the steps that reach it, the action itself, and its verification.
Steps that only reach the action — opening a form, expanding a panel, creating or locating the item to act on — belong to that test. A second operation with its own verification does not.
```

Counting verified operations is what makes the boundary decidable: creating a record you are about to rename is one performed operation but zero verified ones, so it stays inside the test as setup.

Both directions are then pinned by paired examples, since that is where the old wording was ambiguous:

```
Bad: "Open delete dropdown" + "Confirm deletion" — these are ONE test, not two.
Bad: "Create a record, rename it, delete it" — three verified operations, so THREE tests, not one.
Good: "Create a record, rename it and verify the new title" — ONE test; creating is setup, only the rename is verified.
```

Also: all three occurrences of "disposable" dropped from `rules.ts` (now "create its own target", "children created within the scenario itself", "data created by the current scenario"); `planner.ts:352` shortened to `Plan CRUD operations in order: create, read, update, delete.`; and `normal.md`'s arrow chain replaced with one clause.

## Consistency pass

The new Good example legitimizes a third source of test data, which the rules block did not allow — it said propose such scenarios *"only when that data is visible or API preconditions can create it"*, which would forbid a delete test on an empty list even with a visible New button. Added the third clause: *"or the scenario itself creates the record as its setup"*. The phantom-data guard survives, because the create steps must appear in the scenario.

## Verification

- `bun test tests/integration/` — 90 pass / 1 skip / 0 fail
- `bun run format` / `bun run lint` — exit 0
- Three new assertions in `tests/integration/planner.test.ts` guard the wording, including `not.toContain('merge them into one')` and a `not.toContain('disposable')` regression guard.

## Relationship to #168

Taken from #168: the definition sentence, the "separate operations are separate tests" rule, "never rely on another test having run first", the missing-record rule, and the shape of its prompt assertions.

**Not taken:** #168's *"Creating a record is never another test's setup: a test that needs a record says so in the scenario ('an existing item from the list') and the runner prepares the data before the test starts."* That holds only when Fisherman is available. `pilot.ts:748` — with no API, `precondition()` falls through to `checkDataAvailability`, which either lets the tester create via UI or **skips the test**. The traced runs had no `<api_data_preparation>` block at all, so on those runs that rule produces scenarios nothing prepares — reproducing the `f18afa62` failure by design. It also contradicts protectionRule, which forbids deleting pre-existing data, making "an existing item from the list" an illegal target for a delete test.

https://claude.ai/code/session_01B46vQvvc47LJeCmqQxAhme

### DIFF
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -120,7 +120,7 @@ export class Planner extends PlannerBase implements Agent {
       Bad: "Search for X" + "Verify search results" — searching and verifying is ONE test.
       Bad: "Leave field empty" + "Click submit" — that's one negative test, not two.
       Bad: "Create a record, rename it, delete it" — three verified operations, so THREE tests, not one.
-      Good: "Rename existing record and verify the new title" — ONE test; creating is skipped, we asume record already exists, only the rename is verified.
+      Good: "Rename existing record and verify the new title" — ONE test; creating is skipped, we assume record already exists, only the rename is verified.
       You may rely on another test having run first in case we deal with empty state and no relevant data was created yet and we expect another our test creates it
       When the page reports a record is missing or unavailable, it is not a testable surface — plan list-level or recovery behavior instead of operations on that record.${featureDirective}${focusExistingDataDirective}
     </task>
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -114,11 +114,15 @@ export class Planner extends PlannerBase implements Agent {
       Tests must be relevant to the page
       Tests must be achievable from UI
       Tests must be verifiable from UI
-      NEVER split one workflow into multiple tests. Each test must be a complete end-to-end flow.
+      One test verifies ONE business operation: the steps that reach it, the action itself, and its verification.
+      Steps that only reach the action — opening a form, expanding a panel, creating or locating the item to act on — belong to that test. A second operation with its own verification does not.
       Bad: "Open delete dropdown" + "Confirm deletion" — these are ONE test, not two.
       Bad: "Search for X" + "Verify search results" — searching and verifying is ONE test.
       Bad: "Leave field empty" + "Click submit" — that's one negative test, not two.
-      If two scenarios cannot run independently (one requires the other to run first), merge them into one.${featureDirective}${focusExistingDataDirective}
+      Bad: "Create a record, rename it, delete it" — three verified operations, so THREE tests, not one.
+      Good: "Create a record, rename it and verify the new title" — ONE test; creating is setup, only the rename is verified.
+      Never rely on another test having run first: a chained test fails at its first broken step and hides which operations work.
+      When the page reports a record is missing or unavailable, it is not a testable surface — plan list-level or recovery behavior instead of operations on that record.${featureDirective}${focusExistingDataDirective}
     </task>
 
     ${customPrompt || ''}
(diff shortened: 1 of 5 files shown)


src/ai/planner.ts
```diff
@@ -114,11 +114,15 @@ export class Planner extends PlannerBase implements Agent {
...
       Tests must be verifiable from UI
-      NEVER split one workflow into multiple tests. Each test must be a complete end-to-end flow.
+      One test verifies ONE business operation: the steps that reach it, the action itself, and its verification.
+      Steps that only reach the action — opening a form, expanding a panel, creating or locating the item to act on — belong to that test. A second operation with its own verification does not.
       Bad: "Open delete dropdown" + "Confirm deletion" — these are ONE test, not two.
       Bad: "Search for X" + "Verify search results" — searching and verifying is ONE test.
       Bad: "Leave field empty" + "Click submit" — that's one negative test, not two.
-      If two scenarios cannot run independently (one requires the other to run first), merge them into one.${featureDirective}${focusExistingDataDirective}
+      Bad: "Create a record, rename it, delete it" — three verified operations, so THREE tests, not one.
+      Good: "Create a record, rename it and verify the new title" — ONE test; creating is setup, only the rename is verified.
```
DavertMik:
```suggestion
      Good: "Rename existing record and verify the new title" — ONE test; creating is skipped, we asume record already exists, only the rename is verified.
```

src/ai/planner.ts
```diff
@@ -114,11 +114,15 @@ export class Planner extends PlannerBase implements Agent {
...
+      Never rely on another test having run first: a chained test fails at its first broken step and hides which operations work.
```
DavertMik: we may need to do that in case of empty CRUD
```suggestion
      You may rely on another test having run first in case we deal with empty state and no relevant data was created yet and we expect another our test creates it
```

PR #174, opened 2026-09-01
Share prima's envelope tail and give step-file names one owner

DavertMik:
prima's three envelope builders each repeated the `instance`/`status`/`artifacts` tail and depended on an invisible ordering rule — await `saveStatus` before reading a mutable `this.artifacts` field, or silently emit an envelope with no Artifacts section (the exact defect #166 fixed). Now `saveStatus` returns `{hash, artifacts}` and one private `envelopeTail` supplies the shared keys; the field is gone. Adding the next shared envelope field is a one-line change.

The per-step artifact filenames were written twice with nothing enforcing agreement — `writeStepFiles` by string concatenation, the envelope renderer as a hardcoded literal. Both now derive from one `STEP_FILES` constant in `envelope.ts`; the pre-existing test pinning the exact rendered line proves byte-identical output, and a new test asserts the advertised line names every STEP_FILES extension so the writer and the advertisement can no longer drift.

The `status` CLI action re-implemented the command lifecycle inline and diverged (never called `prima.stop()`). `runPrima` gained `{browser, record}` flags and `status` routes through it — still browserless (verified: `status nonexistent` prints a failure envelope, exit 1, Instance line "not running"), now with the same stop/activity handling as every other command. The `config`/`report` unification was deliberately skipped: they diverge on stop() and would need a variation flag.

Note: `writeSnapshot` now calls `getMadeRequests()` (same list; the `getRequests` alias is being removed by open PR #173 — this avoids a conflict between the two PRs). And for the maintainer: the main snapshot still writes `aria.yml` while step files write `.aria.yaml` — same content kind, two spellings, now visible side by side in `envelope.ts`; unifying is a deliberate contract change left undone.

Testing: `bun test boat/prima/tests tests/integration/prima-smoke.test.ts` 146 pass / 1 skip / 0 fail (incl. the new STEP_FILES test), `bun test tests/unit` 1225 pass / 0 fail, biome format + lint clean.

https://claude.ai/code/session_01Jz7wTcYsy68jUUjcWttd1B


boat/prima/src/prima.ts
```diff
@@ -1103,17 +1098,17 @@ export class Prima {
...
 
-  private async writeSnapshot(result: ActionResult): Promise<void> {
-    this.artifacts = writeArtifacts(this.statusDir(), {
+  private async writeSnapshot(result: ActionResult): Promise<ArtifactPaths> {
+    return writeArtifacts(this.statusDir(), {
       aria: result.ariaSnapshot,
       html: await result.combinedHtml(),
       screenshot: result.screenshot,
-      requests: this.bot.requestStore().getRequests(),
+      requests: this.bot.requestStore().getMadeRequests(),
```
DenysKuchma: Need fix

DenysKuchma (changes requested):

PR #177, opened 2026-09-02
Score an overlay from what it blocks, and split Region from Overlay

DavertMik:
Implements `plans/031-overlay-model-cleanup.md`.

An opened panel was classified by **where it sits in the DOM** — anything whose markup was
big enough and whose wrapper happened to be positioned counted as a modal. That missed the
modal this started from (Testomat's suite picker: a real, centred, page-blocking dialog with
no `role="dialog"`, whose subtree is under the 5K floor) and, at the same time, admitted
anything large that merely floated.

Detection now asks what the panel **does to the page**: the share of the viewport it blocks
(1 when it holds the viewport centre, its clipped coverage otherwise) times how much there is
to do inside it (`controls / 3`, capped at 1). Above `0.25` it is an overlay; below, it is a
popover the page can be used around. The size floor stays, but only for in-flow content.

### Model split

`Region` (base, `src/utils/region.ts`) is the general "an area opened on this page" concept and
owns the added-subtree pipeline. `Overlay extends Region` (`src/utils/overlay.ts`) is the
out-of-flow specialisation; `overlay.ts` imports `region.ts` and never the reverse. One
vocabulary replaces four: `detected`→`isModal`, `present`→`isOpen`, `focusArea`/`areaOfInterest`
→ `region`, prompt blocks `<focus_scope>`/`<area_of_interest>` → `<overlay>`/`<region>`,
`probe`→`measure`. No threshold value changed.

### Evidence

Three real-browser fixtures under `test-data/`, driven through Chromium in
`tests/integration/overlay-modal-browser.test.ts` (9 tests):

| Fixture | ARIA says | Measured | Verdict |
|---|---|---|---|
| `testomat_modal.html` — real unmarked modal | heading only, no dialog | fixed, holds centre, 12 controls | overlay `#modal-overlays` |
| `dialog_modal.html` — `role="dialog"` | `dialog "Delete project"` | fixed, holds centre, 3 controls | overlay `#dialog-host` (both paths agree) |
| `cookie_banner.html` — consent bar at the bottom | nothing | fixed, 1280×68 = 9.4% of the viewport, 3 controls | **no region** (score 0.094 < 0.25) |

The cookie banner is deliberately *not* detected: it leaves the page usable, so the Tester
should keep working on the page behind it rather than scoping itself to the bar. Say the word
if you wanted detection there instead — it is one threshold, not a redesign.

### Intended behaviour changes

- Small centred dialogs (confirm/delete prompts) are detected where the 5K floor used to drop them.
- A wide but shallow out-of-flow panel — the notification-panel shape, ~20% coverage with 5
  controls — is now a popover rather than a region. This is the one intended reclassification.
- `z-index > 0` is no longer an out-of-flow signal (it is not a layout property, and it made
  utility-class wrappers count as overlays). CSS `position: fixed|absolute` decides.

### Docs

`docs/superpowers/specs/2026-08-29-region-states-fixes-design.md` §3 describes the band as
"below 5K, a widget, ignored"; after this that holds for **in-flow content only**. Docs were out
of scope for the plan and are untouched.

### Verification

`bun test tests/unit` → 1250 pass, 0 fail (7 new score cases; `origin/main` merged in).
`bun test tests/integration/` → 101 pass, 0 fail. `bun run format` / `bun run lint` → clean.
`bunx tsc --noEmit` adds no new errors over `main` (the repo has 454 pre-existing; CI does not
typecheck).

This needs regression coverage against the Trackly fixture — the score constants are anchored on
shapes, not sites — but I have not labelled it; that call is yours.

### Follow-up: a dialog with no host of its own (`aca02f9`)

Traced from a real session (`TartPalePurple340`), where the tester spent two minutes and a
screenshot finding a confirm button. Two things upstream of the score decided that.

**Scope came from the wrong element.** `measure()` runs on the element that appeared, but `root`,
`html` and `name` were all taken from the diff's grouping container — the nearest element that
already existed above it. For a dialog mounted into a host of its own those coincide. For one
appended into the panel it covers, the layout said "centred, blocking, twelve controls" about the
dialog while the scope handed to the tester named the form behind it, and experience was filed
against that same root.

The diff now records `containerWasEmpty` — it is the only place holding the original DOM — and a
region takes the container as its scope only when it held nothing before; otherwise the element
that appeared is the scope. All three existing fixtures are unaffected (`#modal-overlays`,
`#dialog-host` and `#cookie-host` are all empty before the click), so portal hosts keep their
stable ids instead of degrading to `div.emd-wrapper-target-attachment-center.emd-static`.

**The overlay's markup was cut before its actions.** `HTML_PART_SUBTREE_BUDGET` truncated
head-only, so a picker with a long list lost its footer. Head and tail now share the budget, and
the overlay branch and `collapseHtmlParts` call one `trimSubtree` instead of keeping two copies.

`test-data/inline_picker.html` covers the shape the portal fixtures do not — a centred picker
appended into an occupied panel. Before/after on it:

```
container=main.plan-form  wasEmpty=false  appearedSelector=div.picker-dialog
  before: overlay "Select tests for plan" scope=main.plan-form
  after:  overlay "Select tests for plan" scope=div.picker-dialog
  shown to the AI: 2030 chars, header and both footer buttons present (footer was absent before)
```

`bun test tests/unit` → 1254 pass. `bun test tests/integration` → 103 pass. Format/lint clean.
`origin/main` merged in.

Still open, not touched here: region naming reads only h1–h4, which also decides candidate order
when several subtrees appear in one click.

https://claude.ai/code/session_01U4D6JXK48jVHF6GH35c4ty


src/utils/overlay.ts
```diff
@@ -1,254 +1,167 @@
...
 
-  static fromAria(snapshot: string | null): Overlay {
-    return new Overlay(detectFocusArea(snapshot));
-  }
-
-  static resolve(data: { overlay?: OverlayData | null; ariaSnapshot?: string | null }): Overlay {
-    if (data.overlay) return new Overlay(data.overlay);
+  static resolve(data: { overlay?: RegionData | null; ariaSnapshot?: string | null }): Region {
+    if (data.overlay?.type === 'overlay') return new Overlay(data.overlay);
+    if (data.overlay) return new Region(data.overlay);
```
DenysKuchma, line 34: Could this preserve legacy `type: "modal"` states? They currently fall through to `Region` and lose modal behavior

DenysKuchma (changes requested):

PR #182, opened 2026-09-02
Map a page only when prima research is asked for it

DavertMik:
`prima check` mapped every new url it landed on, and mapped an opening panel on top of that, before it could act. `prima do` was handed a `research` tool it could spend a full model call on in the middle of an instruction. Both now work from the accessibility tree and the page markup, and `prima research` is the one command that maps a page — their help says so, and points at it for a large or unfamiliar page.

### The switch

`ai.agents.researcher.enabled`, read by the Researcher itself:

```javascript
ai: { agents: { researcher: { enabled: false } } }
```

A disabled Researcher answers with the map already recorded for a page and produces none, which closes every route from one place:

| Route | Before | Now |
| --- | --- | --- |
| tester, each new url | full research call | recorded map, else the accessibility tree |
| tester, panel opening | overlay research appended to the map | nothing |
| pilot `ATTACH_UI_MAP` | full research call | recorded map, else nothing |
| `research` tool | full research call | recorded map, else an honest failure |

`prima check` and `prima do` set it for their own run; `prima research` turns it back on, so a host running several prima commands in one process still gets a real map from the command whose whole job is to produce one. An explorbot run leaves it alone and maps pages exactly as before — nothing in `tester.ts` changed.

### A recorded map now joins the context instead of replacing it

`prima do` used to swap the accessibility tree out for the map once a page had been visited three times — carrying off the `[ref=eNN]` element refs the model is told to click by. It also looked the map up under the panel-scoped state hash while research is saved under the page's own, so no map was found while a panel was open. Tree and map are given together now, and `ai.agents.prima.researchAfterVisits` goes with the old behaviour: a map counts from the next command onwards.

### Also here

- The `research` tool returned an empty map as `Successfully researched page`. It now reports that no map is available and points at `context()`.
- A recorded map handed to a run is reported on the `research` data frame like a freshly produced one, so a host watching over `--ws` sees the map the run is acting on.
- Prima's help texts move out of `cli.ts` into `boat/prima/src/help.ts`, leaving the CLI as command wiring.

### Testing

`bun test tests/integration/` 103/103 (also run in reversed file order) and `bun test tests/unit` 1281/1281.

Not covered by tests: no scenario here exercises a `check` run against a real app. Worth a regression label if you want that before merge.

https://claude.ai/code/session_01PbuGGh8w9RJJhbCQt7iLHQ


boat/prima/src/prima.ts
```diff
@@ -583,7 +582,14 @@ export class Prima {
   }
 
   private async loadConfig(): Promise<ExplorbotConfig> {
-    return ConfigParser.getInstance().loadConfig({ config: this.options.config, path: this.options.path, baseUrl: this.configBaseUrl() });
+    const config = await ConfigParser.getInstance().loadConfig({ config: this.options.config, path: this.options.path, baseUrl: this.configBaseUrl() });
+    const ai = config.ai;
+    if (ai) {
+      ai.agents ??= {};
+      ai.agents.researcher ??= {};
+      ai.agents.researcher.enabled ??= this.options.command === 'research';
```
DenysKuchma: For multiple Prima commands in one process. After `do/check` sets `enabled` to false, a later `research` command cannot turn it back on because `??=` preserves false

DenysKuchma (changes requested):

boat/prima/src/prima.ts
DavertMik: Prima runs 1 command one shot 
So should be ok

Issue #184, opened 2026-09-03
Doc Collector Subagent


DavertMik:
We need to create a subagent which will pick up documentation from docbot (via frontmatter) and existing docs from markdown (using ripgrep or grep tools). 

- Planner should query this subagent to find relevant data and it should provide this data back to planner
- There must be configurable option in planner to propose test ideas from documentation or web. And combination like 70/30 docs to web discovered tests
- Enabling this subagent is configurable, as it requires documentation already pre-fetched

```
Plans
    ...knowledge collector agent 
        > Find all documented plan capabilities
            which rg
            rg
                > ....
                > ....
            grep
                > ...
        > docbot
            frontmatter ...

config: .... docs: 70%, web: 30%        

```

Proposed names:

- Lookout
- **Scout**
- Spotter
- Observer
- Lighthouse
- **Beacon**
- **Periscope**
- Spyglass
- Crow's Nest
- **Sonar**
- Radar
- **Cartographer**

PR #189, opened 2026-09-04
Let the Planner draw scenarios from collected documentation

DenysKuchma:
## Summary

 This PR adds Scout — a new optional agent that retrieves documentation for the Planner. https://github.com/testomatio/explorbot/issues/184

  When planning a page, Scout searches the collected documentation (the docbot application spec bundle and any extra markdown folders configured via `ai.agents.scout.dirs`) using ripgrep or grep, reads the relevant files and reports a digest with sources. The Planner receives it as a `<docs_context>` block next to the existing `<application_spec>` and grounds part of its scenarios in it — the share is controlled by `ai.agents.planner.docsWeight` (0-100, default 70). Pages already injected for the current URL are excluded, so the two blocks never duplicate each other.
 Hand-written notes from `scout.dirs` go into Scout's context directly, so they reach the Planner reliably even on cheap models.

  Scout is off by default: enable it with `ai.agents.scout.enabled: true` (it requires documentation collected beforehand, e.g. via `explorbot docs collect`). If rg/grep are not installed, it falls back to an in-process scan. Covered by unit and integration tests and verified with live runs against a real docbot corpus.


src/ai/scout/search.ts
```diff
@@ -0,0 +1,142 @@
...
+import { resolve } from 'node:path';
+import { loadMarkdownFiles } from '../../utils/markdown-files.ts';
+
+export const MAX_HITS = 50;
+const HIT_LINE_CLAMP = 200;
+const HIT_PATTERN = /^(.+?):(\d+):(.*)$/;
+
+let cachedEngine: SearchEngine | null = null;
+
+export async function searchMarkdown(dirs: string[], pattern: string, opts: SearchOptions = {}): Promise<SearchHit[]> {
```
DavertMik: No, this is overcomplicated implementation;
Agents already good trained to call grep/rg for scanning

What we need - detect which one available, throw if no is available
Provide bash tool to scout agent on a working directly 
And tell it to use rg/grep for scanning

So the only tools needed are bash+readFile which we already have in Captain

So mostly scout should invest in good system prompt and allow freedom of exploration
We can use agenticModel for it as I don't assume scanning files with grep will consume lots of tokens
but agenticModel should natively work with filesystems

DavertMik (changes requested):
Let's simplify implementation and let model descide rg/grep to be called

DenysKuchma:
- Removed the custom search engine and search/report tools.
- Scout now uses Captain’s bash + readFile tools and the configured agenticModel.
- Added rg/grep detection with an explicit error when neither is available.
- Removed direct hand-written documentation injection; Scout now explores the corpus only through tools.
- Fixed a Bun/WSL race by starting Scout after browser research completes.
- Added corpus access restrictions and tests for enable/disable, docsWeight, scanner fallback, grounding, and execution order.

DavertMik (approved):

PR #192, opened 2026-09-07
Keep the API proof Pilot already read, and let it reload to check persistence

DavertMik:
Session `HushedRunningEmerald179` set a milestone on a suite, could not verify it, and stopped. Trace `dd01571ee44f24d113184e54294de7bc`, ~4.5 minutes, 264 observations. The milestone was saved and retained the whole time — the run had proof of it twice and threw it away both times.

## The API answer was discarded

At 02:53:49 Pilot called `askApi("… which milestone is currently assigned?")`. Fisherman ran `getEndpointSpec` + `GET /api/…/suites/c44bb77f` and got a **200 carrying `milestones:[{id:"new",title:"Milestone"}]`**. Then three consecutive model calls errored, `runSession`'s `catch` stopped the loop, and `finished` was never set — so `getResult()` fell through to `synthesizeResult(haul, declaredDone=false, readOnly=true)`, whose `success: declaredDone && succeeded.length > 0` is false by construction. Pilot received:

```json
{"answered": false, "reason": "Stopped before finishing: 1 requests, 1 successful reads, 0 failed"}
```

The read was in the haul. Only the step that phrases it failed, and the whole answer went with it.

A read-only lookup that ends without an explicit answer now returns the responses it actually read (last 3, `BODY_PREVIEW_LIMIT` each). An explicit `stop` still wins — it sets `result` and short-circuits `getResult()` — so a deliberate "these endpoints cannot answer this" is not converted into a false positive.

## The reload was vetoed

`buildResetSystemPrompt` knew reset only as a destructive redo — *"Reset is DESTRUCTIVE… Unnecessary resets create duplicate data"*, `PRIORITY 1) Successful side effects → almost never allow reset`. Nothing said reset merely re-navigates and writes nothing, so an expectation like *"The milestone remains assigned after the page reloads"* could never get its reload once the write landed.

Pilot vetoed at 02:53:28 with *"post-reload visibility was already verified"* — that verification had passed on a **different suite** the tester drifted into. Forty-four seconds later its own finish verdict asked for exactly what it had refused: *"Reload the current suite, then verify the same milestone chip remains visible."*

`allow` now covers a persistence expectation while no reset has been taken this run. It keys off the reset count the review is already given verbatim (`Previous reset count: N`), not off "did a reload happen" — `buildTaskContext` does not expose `startUrl` to the reset review, so Pilot cannot tell "reopened a record" from "returned to the start page", which is the inference it already got wrong here.

## Changes

| File | Change |
|---|---|
| `src/ai/fisherman/tools.ts` | `synthesizeResult` returns the successful reads as the answer for a read-only lookup |
| `src/ai/fisherman/tools.ts` | `askApi` description: reading a record back also settles whether a change was *saved*, not only what exists beforehand |
| `src/ai/fisherman/tools.ts`, `src/ai/fisherman.ts` | both answer log lines truncated to 200 — the answer is a body now, the log pane is not |
| `src/ai/pilot.ts` | `buildResetSystemPrompt`: reset writes nothing; `allow` covers persistence expectations; `continue` narrowed to "observable on the current page"; `PRIORITY 1` scoped to re-observe vs repeat |
| `tests/integration/fisherman.test.ts` | aimock case reproducing the trace: GET succeeds, next model turn fails, the answer still carries the body |

No tools were added to `reviewDecision`. `createAskApiTool` already files the answer via `task.addNote`, and the verdict reads `<notes>` — with this fix the 02:53:49 answer would have been in front of the 02:54:12 verdict that said *"retention not explicitly verified"*.

## Trade-offs

`reviewResetDecision` injects guidance only on `continue`, not on `allow`, so a tester that reloads arrives with no steer to observe rather than redo — the duplicate-write risk moves from "the reload never happens" to "the tester repeats the flow after it"; `resetCount` and the `fail` branch stay the only brake.

A lookup that ends without an explicit answer hands Pilot raw JSON instead of a sentence, and `addNote` carries it into the plan file and the Analyst's report. It only fires when the phrasing step failed, and it is the evidence channel, so it stays.

## Verification

- `bun test tests/unit/` — 1315 pass, 0 fail
- `bun test tests/integration/` — 111 pass, 1 fail (`prima-smoke`, needs a live playwright-cli session; **fails identically on `main`**)
- `bunx tsc --noEmit -p tsconfig.json` — no new errors in changed files

Not replayed against the live scenario — this touches Pilot's judgement, so it likely wants the `regression` label.

https://claude.ai/code/session_017apAVyv97GNLeM7pSvZTn2

### DIFF
--- a/src/ai/fisherman/tools.ts
+++ b/src/ai/fisherman/tools.ts
@@ -7,11 +7,13 @@ import type { RequestStore } from '../../api/request-store.ts';
 import { extractEndpointDefinition } from '../../api/spec-reader.ts';
 import type { Test } from '../../test-plan.ts';
 import { tag } from '../../utils/logger.ts';
+import { truncate } from '../../utils/strings.ts';
 import { isDynamicSegment } from '../../utils/url-matcher.ts';
 import type { Fisherman } from '../fisherman.ts';
 import type { RequestHaul } from './request-haul.ts';
 
 const BODY_PREVIEW_LIMIT = 2000;
+const READS_IN_ANSWER = 3;
 
 export function createFishermanTools(apiClient: ApiClient, requestStore: RequestStore, haul: RequestHaul, opts: { spec?: any; baseEndpoint?: string; readOnly?: boolean }) {
   const readOnly = opts.readOnly === true;
@@ -222,6 +224,7 @@ export function createAskApiTool(fisherman: Fisherman | null, task: Test) {
         Ask what data already exists, changing nothing.
         Ask a question about existing records: which ones are there, what they are called, whether a particular one exists.
         Use it before precondition() to see whether suitable data is already available, and whenever a step needs the exact name or id of a record that is already there.
+        Use it after a change the page does not show as well: reading the record back is what settles whether the change was saved.
         It never creates, edits or deletes anything — precondition() does that.
       `,
       inputSchema: z.object({
@@ -242,7 +245,7 @@ export function createAskApiTool(fisherman: Fisherman | null, task: Test) {
         }
 
         task.addNote(`Asked API: ${question} — ${result.summary}`);
-        tag('success').log(`Ask API: ${result.summary}`);
+        tag('success').log(`Ask API: ${truncate(result.summary, 200)}`);
         return { answered: true, answer: result.summary };
       },
     }),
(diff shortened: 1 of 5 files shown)


src/ai/fisherman/tools.ts
```diff
@@ -222,6 +224,7 @@ export function createAskApiTool(fisherman: Fisherman | null, task: Test) {
         Ask what data already exists, changing nothing.
         Ask a question about existing records: which ones are there, what they are called, whether a particular one exists.
         Use it before precondition() to see whether suitable data is already available, and whenever a step needs the exact name or id of a record that is already there.
+        Use it after a change the page does not show as well: reading the record back is what settles whether the change was saved.
```
DavertMik:
```suggestion
```

PR #195, opened 2026-09-08
Stop the click tool accepting a coordinate click as a locator

DavertMik:
Found while looking at sessions `AlertMeanAmethyst509` and `JustParentalOrange422`, where the tester kept acting on controls it had never pressed.

## What went wrong

`I.clickXY(0, 0)` cannot fail. With a numeric first argument, CodeceptJS's Playwright helper goes straight to `page.mouse.click(0, 0)` — no `_locateElement`, no `assertElementExists`, no bounding box. The `click` tool returns on the first command that doesn't throw, so a coordinate click at the end of a fallback list converts "none of my locators matched" into `success: true`.

In `AlertMeanAmethyst509` that happened twice. The model offered four real locators for "Load more" plus `I.clickXY(0, 0)`; the four failed, the corner click "succeeded", and the tool reported it as the working locator. The tester then recorded *"New rows appended to list per page diff"* as evidence the feature worked. Eleven seconds earlier `xpathCheck` had told it the truth — `hit: "offscreen"`, *"Element exists in DOM but its interaction point is outside the viewport"* — and the success report overrode it.

The model didn't invent the idea. It was in the prompts three times: as the first dismiss step in `unexpectedPopupRule`, as "If nothing works, use I.clickXY(x, y) as last resort" in `actionRule`, and as item 5 in the click tool's own schema.

## Changes

- The click tool refuses `I.clickXY` commands and names the alternatives: `visualClick`, which locates the target in a screenshot before clicking it, and `form()` for a deliberate coordinate click.
- Its schema drops coordinates from the fallback ladder.
- `unexpectedPopupRule` leads with Escape, then Cancel and Close. **Clicking outside a layer to dismiss it is kept** — it's a real pattern, and the harm was never the coordinate click itself, it was one sitting in a *locator* list where "didn't throw" reads as "found the element". As a deliberate act it goes through `form()`, which already gates on an observable page change.
- A click that runs without error but leaves URL, ARIA and HTML untouched is now reported as failed, reusing the `hasObservablePageChange` gate the `form` tool already applies. The suggestion points at `xpathCheck`, which reports whether an element is covered or offscreen.

## Coverage, honestly

`hasObservablePageChange` now also counts `pageDiff.requests`. Network calls are already filtered to same-origin xhr/fetch, so they are a clean signal of the app doing work — without this, a Save that stores a record without redrawing anything came back failed and the model would retry it into a duplicate. The change is in the shared helper, so `form()` gets it too. Neither `clickXY(0, 0)` in the session produced any request, so the fix above is unaffected.

The no-change gate catches the second bad click in that session — the pagination one, whose diff was completely empty. It would **not** have caught the first: that diff carried `messages: ["And 31 items after"]` and an `htmlParts` entry, both of which count as observable change. Those 31 rows were a real change the app made on its own, and the tool has no way to know the click didn't cause them. That call is fixed only by removing coordinates from the click tool.

## Not touched

- **Driller / Researcher** still use coordinate clicks. `researcher/deep-analysis.ts` and `researcher.ts` run them through `action.attempt()` directly, not this tool, so they're unaffected.
- **Overlay detection.** Neither session's panel was ever detected: `overlay.ts` finds a region either as an added subtree in a diff or by an ARIA `dialog` role, and both tests began with `I.amOnPage` straight onto a route whose panel was already rendered — no opening to diff, and zero `dialog` roles in either initial snapshot. That cost `JustParentalOrange422` five attempts and 65 seconds on a "New" button duplicated between the panel and the page behind it; the same selector worked instantly once Cancel closed the panel. Worth fixing separately.
- **Click-to-focus.** `focused_element_actions` tells the model to click a field before `type()`. Focus is captured on `ActionResult` but never compared in `diff()`, so a click whose only effect is moving focus is now reported failed. Fixing it means plumbing `focusedElement` into `pageDiff`; called out here rather than widened into this PR.
- **`action-result.ts:563`**, which fills `pageDiff.htmlParts` with a region snapshot carrying `added: []` / `removed: []`. That entry masks the no-change check. A newly opened region genuinely is a change, so this is only wrong when the detection itself was a false positive — follow-up, not this PR.

## Tests

`tests/unit/click-coordinates.test.ts` — four cases: a bare coordinate command, coordinates as the last fallback of a locator list, and both no-change shapes (identical state id, and a diff with no url/aria/html change). Five cases including a click answered only by an API call. Unit suite 1320 pass / 0 fail; integration 110 pass / 0 fail.

https://claude.ai/code/session_017uQPf49i79cxZS7ZQseDSS

### DIFF
--- a/src/ai/rules.ts
+++ b/src/ai/rules.ts
@@ -238,10 +238,10 @@ export const unexpectedPopupRule = dedent`
   If buttons are disabled unexpectedly, check if a popup is blocking interaction or if required form fields are empty.
 
   Dismiss strategy (try in order):
-  1. I.clickXY(0, 0) — click outside the popup to close it
-  2. I.pressKey('Escape') — press Escape to dismiss
-  3. I.click('Cancel') — click Cancel button if present
-  4. I.click({ role: 'button', text: 'Close' }) — click X/close button if present
+  1. I.pressKey('Escape') — press Escape to dismiss
+  2. I.click('Cancel') — click Cancel button if present
+  3. I.click({ role: 'button', text: 'Close' }) — click X/close button if present
+  4. Click a point outside the layer — a deliberate coordinate click, so run it through form(), not click()
   </unexpected_popup_rule>
 `;
 
@@ -335,7 +335,7 @@ export const actionRule = dedent`
   Prefer text/ARIA locators with context over complex CSS/XPath selectors.
   For inline create/edit flows, after filling a field verify it contains the value, then confirm using the nearest explicit button/link, an adjacent icon-only confirm control in the same row/form, or Enter if the field remains focused.
   If locator doesn't work, try CSS or XPath locators.
-  If nothing works, use I.clickXY(x, y) as last resort.
+  If nothing works, use visualClick() — it locates the target in a screenshot before clicking it.
 
   When a click result reports several matches, pick one from its numbered list by position rather than guessing a new locator.
   Reuse the same locator with step.opts({ elementIndex: N }) as the LAST argument. N is the "Element N" number.
--- a/src/ai/rules.ts
+++ b/src/ai/rules.ts
@@ -241,7 +241,7 @@ export const unexpectedPopupRule = dedent`
   1. I.pressKey('Escape') — press Escape to dismiss
   2. I.click('Cancel') — click Cancel button if present
   3. I.click({ role: 'button', text: 'Close' }) — click X/close button if present
-  4. Click a point outside the layer — a deliberate coordinate click, so run it through form(), not click()
+  4. I.clickXY(x, y) on the backdrop beside the layer — run it through form(), not click(), and avoid page corners, which usually hold their own links
   </unexpected_popup_rule>
 `;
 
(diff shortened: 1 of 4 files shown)


src/ai/rules.ts
```diff
@@ -238,10 +238,10 @@ export const unexpectedPopupRule = dedent`
...
 
   Dismiss strategy (try in order):
-  1. I.clickXY(0, 0) — click outside the popup to close it
-  2. I.pressKey('Escape') — press Escape to dismiss
-  3. I.click('Cancel') — click Cancel button if present
-  4. I.click({ role: 'button', text: 'Close' }) — click X/close button if present
+  1. I.pressKey('Escape') — press Escape to dismiss
+  2. I.click('Cancel') — click Cancel button if present
+  3. I.click({ role: 'button', text: 'Close' }) — click X/close button if present
+  4. I.clickXY(x, y) on the backdrop beside the layer — run it through form(), not click(), and avoid page corners, which usually hold their own links
```
DavertMik:
```suggestion
  4. I.clickXY(0, 0) via form() tool and check if page diff changed
```

src/ai/tools.ts
```diff
@@ -66,7 +66,7 @@ export function createCodeceptJSTools({ explorer, stateManager }: ToolDeps, task
           2. I.click(ARIA, container) - e.g. I.click({"role":"button","text":"Save"}, ".modal")
           3. I.click(CSS, container) - e.g. I.click("#btn", ".modal")
           4. I.click(CSS) or I.click(XPath) - when locator already includes context (ID, XPath)
-          5. I.clickXY(x, y) - coordinates fallback
+          Every command must name an element. Coordinates are not a locator — use visualClick() when you can see the target but cannot address it.
```
DavertMik:
```suggestion
```

PR #197, opened 2026-09-08
apibot: endpoint patterns, the command pattern, and an init that writes a usable config

DavertMik:
## The bug

apibot could not find a real path in its own spec. Specs write paths as templates, so `explore /api/v2/zyntra-don-t-touch-cloned/tests` against `/api/v2/{project_id}/tests` matched nothing and the run died before planning:

```
Failed: Endpoint "/api/v2/zyntra-don-t-touch-cloned/tests" not found in spec.
Available: /api/v2/{project_id}/info, /api/v2/{project_id}/tests, ...
```

`matchesEndpoint` compared strings. It is segment matching now: a template segment stands for any value, a literal segment wins over a template, and the paths below the match come along. So one collection hands the planner its create, read, update and delete. When nothing resolves it falls back to the old literal prefix match, so a wrong endpoint still gets the same error rather than the whole spec.

## Endpoint patterns

`api explore` takes a pattern instead of a single endpoint.

```bash
npx explorbot api explore /users                # one endpoint, every planning style
npx explorbot api explore '/projects/acme/*'    # every collection of one project
npx explorbot api explore / --endpoint https://api.example.com/v2/acme
```

`*` stands for one segment. Matches fold into their collection, so `/users/{id}` and `/users/{id}/posts` explore as `/users` and the deeper paths ride along in the spec lookup. `/` takes everything, with path parameters coming from the base endpoint.

One endpoint behaves exactly as before, every style. Several collections share the styles, one per collection, so covering a whole API is one plan per collection rather than one per style. Parameters the pattern cannot fill are named: a collection or two get skipped with a warning, and a run with nothing left stops and says which parameter is missing instead of sending requests to a literal `{project_id}`.

Against the Testomat.io v2 spec: `/tests` and `/tests/*` both resolve to one endpoint, `/<project>/*` to 19, skipping `/analytics/tests/{kind}` and `/analytics/stats/{kind}`.

## The command pattern

Every apibot command was a closure inside `cli.ts`, so none of them could be tested. They are `BaseCommand` classes now, one file each under `boat/api-tester/src/commands/`, and `cli.ts` drops from 325 lines to 146: build the bot, run the command, pick an exit code.

`BaseCommand` takes its collaborator as a type parameter, `BaseCommand<T = ExplorBot>`, so the thirty-five explorbot commands compile untouched and apibot's extend `BaseCommand<ApiBot>` through a small `ApiCommand` base that names the field `bot`. `init` stays a plain exported function, as `explorbot init` is, since there is no bot to hang it on.

The duplicated Curler invocation that `explore` and `test` both carried is now `ApiBot.runTest`.

## Init

The generated config named a model nobody here uses, behind an import of a package the project never installs:

```js
import { openai } from '@ai-sdk/openai';
export default {
  ai: { model: openai('gpt-4o') },
  api: { baseEndpoint: '...', headers: { /* 'Authorization': 'Bearer <token>' */ } },
  dirs: { output: 'output', knowledge: 'knowledge' },
};
```

Models now come from `ConfigParser.recommendedModels()` as `provider/model-id` strings, the way `explorbot init` writes them. The commented-out `bootstrap`/`teardown` blocks, the empty `headers` block and the `dirs` block that only restated the defaults are gone, a `.env` with the provider keys is written alongside, and the file is `apibot.config.js`, which is what the loader looks for first and what its own "config missing" error names. New `--provider`, `--endpoint` and `--spec` options skip the questions.

## Tests

`tests/unit/spec-reader.test.ts` (15), `tests/unit/apibot-explore.test.ts` (3, covering the style cycling the CLI could not reach), `tests/unit/apibot-init.test.ts` (5, asserting the template stays clean). Full unit suite: 1361 pass.

## Notes

- `plan` and `test` still take a single endpoint. `*` there stays undefined.
- No cap on a wide run, by choice.
- Chief already dedupes scenarios across a session, so the collections in one run will not repeat an identical scenario sentence.

https://claude.ai/code/session_01CZkcbkiidWcYj4Cpo69pGL

### DIFF
--- a/boat/api-tester/src/commands/init-command.ts
+++ b/boat/api-tester/src/commands/init-command.ts
@@ -1,0 +1,116 @@
+import { existsSync, mkdirSync, writeFileSync } from 'node:fs';
+import path from 'node:path';
+import chalk from 'chalk';
+import { envTemplate, modelLines } from '../../../../src/commands/init-command.ts';
+import { missingModelRoles } from '../../../../src/config.ts';
+import { log, tag } from '../../../../src/utils/logger.ts';
+
+export async function runInit(options: InitOptions): Promise<void> {
+  const provider = options.provider || 'openrouter';
+  const originalCwd = process.cwd();
+
+  if (options.path) {
+    const dir = path.resolve(options.path);
+    mkdirSync(dir, { recursive: true });
+    process.chdir(dir);
+    log(`Working in directory: ${dir}`);
+  }
+
+  const configPath = path.resolve('apibot.config.js');
+  if (existsSync(configPath) && !options.force) {
+    log(`Config file already exists: ${configPath}`);
+    log('Use --force to overwrite existing file');
+    process.exit(1);
+  }
+
+  const answers = await ask(options);
+  if (!answers.baseEndpoint) {
+    tag('error').log('Base endpoint is required.');
+    process.exit(1);
+  }
+
+  writeFileSync(configPath, configTemplate(provider, answers.baseEndpoint, answers.spec), 'utf8');
+  log(`Created config file: ${configPath}`);
+
+  const envPath = path.resolve('.env');
+  if (!existsSync(envPath)) {
+    writeFileSync(envPath, `${envTemplate(provider)}\n`, 'utf8');
+    log(`Created env file: ${envPath}`);
+  }
+
+  mkdirSync('output', { recursive: true });
+  mkdirSync('knowledge', { recursive: true });
+
+  if (answers.knowledge) {
+    const knowledgePath = path.resolve('knowledge', 'general.md');
+    writeFileSync(knowledgePath, `---\nendpoint: "*"\n---\n${answers.knowledge}\n`, 'utf8');
+    log(`Created knowledge file: ${knowledgePath}`);
+  }
+
+  const missing = missingModelRoles(provider);
+  if (missing.length) {
+    tag('warning').log(`No recommended ${missing.join(' and ')} for ${provider} — set the model ids in ${configPath}`);
+  }
+
+  log('');
+  log('Next steps:');
+  log('1. Add your provider API key to .env');
+  log('2. Describe the API so the plans match it');
+  tag('substep').log(chalk.yellow(`${options.prefix} know /users "CRUD endpoint for user management"`));
+  log('3. Plan and run tests for one endpoint');
+  tag('substep').log(chalk.yellow(`${options.prefix} explore /users`));
+
+  if (process.cwd() !== originalCwd) process.chdir(originalCwd);
+}
+
+function configTemplate(provider: string, baseEndpoint: string, spec: string): string {
+  let specLine = '';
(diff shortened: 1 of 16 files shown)


boat/api-tester/src/commands/init-command.ts
```diff
@@ -0,0 +1,116 @@
...
+    return { baseEndpoint: options.baseEndpoint, spec: options.spec || '', knowledge: '' };
+  }
+
+  const rl = await import('node:readline');
+  const iface = rl.createInterface({ input: process.stdin, output: process.stdout });
+  const question = (text: string): Promise<string> => new Promise((resolve) => iface.question(text, (answer: string) => resolve(answer.trim())));
+
+  log('Apibot — API Testing Tool Setup\n');
+  const baseEndpoint = await question('Base API endpoint (e.g. https://api.example.com/v1): ');
+  const spec = await question('OpenAPI spec file or URL (Enter to skip): ');
```
DenysKuchma: Spec cannot be optional here ApiBot.start() always calls validateSpecs()
Skipping it generates a config that immediately fails on plan or explore

src/api/spec-reader.ts
```diff
@@ -166,6 +195,75 @@ function stripBasePath(specPath: string, basePath: string): string {
...
+
+function segmentsMatch(segments: string[], wanted: string[]): boolean {
+  return wanted.every((want, i) => want === '*' || segments[i] === want || segments[i].startsWith('{'));
+}
+
+function toCollection(specPath: string, specPaths: string[]): string {
+  const segments = toSegments(specPath);
+  for (let i = 1; i < segments.length; i++) {
+    const prefix = `/${segments.slice(0, i).join('/')}`;
+    if (specPaths.includes(prefix)) return prefix;
```
DenysKuchma, line 243: This selects the first existing ancestor and can collapse nested collections too far
For example, /projects/acme/* can resolve only to /projects, losing the matched tests and suites collections

src/api/spec-reader.ts
```diff
@@ -61,6 +61,35 @@ export function extractEndpointDefinition(schema: any, endpoint: string, baseEnd
...
+  const basePath = toBasePath(baseEndpoint);
+  const normalized = Object.keys(schema.paths).map((specPath) => stripBasePath(specPath, basePath));
+  const matched = normalized.filter((specPath) => matchesPattern(specPath, pattern));
+
+  if (!matched.length) {
+    throw new Error(`Endpoint "${pattern}" not found in spec. Available: ${listNormalizedPaths(schema, basePath)}`);
+  }
+
+  const roots = [...new Set(matched.map((specPath) => toCollection(specPath, normalized)))];
+  const resolved = roots.map((root) => fillParameters(root, pattern));
```
DenysKuchma: Different literal and templated roots can become the same endpoint after parameter substitution
Deduplicate after fillParameters() to avoid exploring the same endpoint twice

DenysKuchma (changes requested):

PR #203, opened 2026-09-10
Let the Planner see how its tests ended

DenysKuchma:
## Summary

 The Planner planned blind. The session test list carried only scenario titles, successful flows were the only experience it ever saw, and one of its own rules asked for "data persistency after page reload" outcomes - so the same doomed patterns were proposed on page after page. In the last night run (13 pages, 57 tests) seven "verify it persists after reload" scenarios produced three failures and four unfinished tests, and zero passes

### DIFF
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -359,7 +359,10 @@ export class Planner extends PlannerBase implements Agent {
       You can't test emails, database, SMS, or any external services.
       Suggest scenarios that can be potentially verified by UI.
       Focus on error or success messages as outcome.
-      Focus on URL page change or data persistency after page reload.
+      Focus on URL page change as outcome.
+      Every expected outcome must name what the page shows when it happens: a message, a list change, a control state, or a URL change.
+      Persistency after a reload counts only when the scenario names the on-screen evidence that will show it — a visible marker, a list entry, a stored setting shown in the interface.
+      An outcome the page cannot display — a clipboard write, an internal state with no visible trace, "the operation succeeds" with nothing shown — is not verifiable and must not be an expected outcome.
       If there are subpages (pages with same URL path) plan testing of those subpages as well
       Plan CRUD operations in order: create, read, update, delete.
       Do not invent specific route names, success messages, validation texts, badge counts, or welcome messages unless they are visible in research, visited pages, or prior observed flows.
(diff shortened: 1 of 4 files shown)


src/ai/planner.ts
```diff
@@ -374,6 +377,7 @@ export class Planner extends PlannerBase implements Agent {
       If the list is empty or no concrete item names are visible, do not invent "known" or "existing" items. Prefer empty-state, no-match search, clear-search, or read-only list behavior scenarios.
       Search, filter, sorting, tab, and list scenarios must start from a stable page where those controls are visible; avoid transient create/edit/new URLs unless the scenario tests that form.
       For option values and list items, use only visible or previously observed data; do not add create/update/delete setup unless the user explicitly requests that workflow.
+      A scenario that edits, deletes, reassigns, or reconfigures a record must act on a record the scenario itself creates as its setup — that setup is required, not optional, because records the app already had are protected from mutation. Acting on a pre-existing record makes the scenario unrunnable, however visible the record is.
```
DavertMik: leads to the idea that every record must be created
which is not optimal, we will end up scenarios like:

create & edit it 
create & change title
create & ...

not optimal
creation can be done quickly by pilot+fisherman, or picked by pilot

I don't think data preparation is part of scenario planning.

src/ai/planner.ts
```diff
@@ -359,7 +359,10 @@ export class Planner extends PlannerBase implements Agent {
       You can't test emails, database, SMS, or any external services.
       Suggest scenarios that can be potentially verified by UI.
       Focus on error or success messages as outcome.
-      Focus on URL page change or data persistency after page reload.
+      Focus on URL page change as outcome.
+      Every expected outcome must name what the page shows when it happens: a message, a list change, a control state, or a URL change.
```
DavertMik: we can get a very strict list of expectations
howver planner doesn't see subpages

src/ai/planner/session-dedup.ts
```diff
@@ -18,7 +18,15 @@ export function WithSessionDedup<T extends Constructor>(Base: T) {
...
-          lines.push(`${plan.url || '/'} | ${test.style || 'default'} | ${test.scenario}`);
+          const lastNote = Object.values(test.notes)
+            .filter((note) => note.message)
+            .pop();
+          const outcome = test.result || (lastNote ? 'unfinished' : 'pending');
+          let line = `${plan.url || '/'} | ${test.style || 'default'} | ${outcome} | ${test.scenario}`;
+          if ((outcome === TestResult.FAILED || outcome === 'unfinished') && lastNote) {
+            line += ` — ${lastNote.message.slice(0, 140)}`;
+          }
+          lines.push(line);
```
DavertMik: should be moved to function and unit tested
i really don't understand what happens here

also please avoid ternary operators!

DavertMik (changes requested):

PR #205, opened 2026-09-11
 Anchor planning to stable pages, not single records

DenysKuchma:
## Summary
 
 When a record dies mid-run, everything planned on it dies with it. In the last night run the sub-page picker visited the edit URL of a record an earlier test had deleted, the planner built five scenarios on that page, and all of them stopped on a 404 - about a quarter of the night's failed tests trace back to this shape.

  Two general rules close it:

  - The sub-page picker prefers stable pages over record forms: a page that edits or configures one specific record is a step inside a feature, not a feature's planning base, and the record it addresses may already be gone — pick the page that lists the collection instead.
  - Scenarios anchor to the stable page holding a collection, never to the URL or ID of one specific record; the steps locate the record from that page.

### DIFF
--- a/src/ai/planner.ts
+++ b/src/ai/planner.ts
@@ -373,6 +373,7 @@ export class Planner extends PlannerBase implements Agent {
       For scenarios that act on existing items or search/filter by existing values, use only item names or values visible in research, visited pages, or prior observed flows.
       If the list is empty or no concrete item names are visible, do not invent "known" or "existing" items. Prefer empty-state, no-match search, clear-search, or read-only list behavior scenarios.
       Search, filter, sorting, tab, and list scenarios must start from a stable page where those controls are visible; avoid transient create/edit/new URLs unless the scenario tests that form.
+      A scenario anchors to a stable page that holds the collection — never to the URL or ID of one specific record, which another test may have deleted by the time this one runs; steps locate the record from the stable page instead.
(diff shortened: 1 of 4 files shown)


src/ai/planner.ts
```diff
@@ -373,6 +373,7 @@ export class Planner extends PlannerBase implements Agent {
       For scenarios that act on existing items or search/filter by existing values, use only item names or values visible in research, visited pages, or prior observed flows.
       If the list is empty or no concrete item names are visible, do not invent "known" or "existing" items. Prefer empty-state, no-match search, clear-search, or read-only list behavior scenarios.
       Search, filter, sorting, tab, and list scenarios must start from a stable page where those controls are visible; avoid transient create/edit/new URLs unless the scenario tests that form.
+      A scenario anchors to a stable page that holds the collection — never to the URL or ID of one specific record, which another test may have deleted by the time this one runs; steps locate the record from the stable page instead.
```
DavertMik: I think we need to say it only once (no duplication among lines):

Do not use record IDs or uniq titles inside test plans 
Pilot will pick the correct record name
Planner plans only scenario, it should explain which record is needed but not pick exact one

Exception: when we see limited list of records

DavertMik (changes requested):
- One short prompt change preferable
- Do not name 'collections' as it is too explicit assumption 
- Ask not to name records or ids in plans
- records can be named only when collection is known and limited

PR #220, opened 2026-09-16
Give every site its own config extending the global one

DavertMik:
A global installation had a single config for every site, so models, browser, and reporter settings could not differ per app. There was also no obvious link between a config and the site it applied to.

Each site folder now holds its own `explorbot.config.js` that extends `~/.explorbot/config.js`:

```
~/.explorbot/
├── config.js                    # shared by every site
├── .env
└── sites/
    └── app.example.com/
        ├── explorbot.config.js  # this site's settings
        ├── site.json
        ├── knowledge/
        ├── experience/
        └── output/
```

The two are merged section by section and the site wins, so a site that overrides `ai.model` keeps the global `ai.visionModel`. Use it to give a slow or unusual app a stronger model or a visible browser without changing how every other site runs.

## web.url is required

The site config must declare `web.url` naming its own site, so the file states what it configures rather than meaning something only by where it sits. A missing or mismatched URL stops the run and names the file:

```
Site config declares a different site.
  ~/.explorbot/sites/app.example.com/explorbot.config.js
  web.url:  https://staging.example.com
  site:     https://app.example.com

Fix web.url, or explore https://staging.example.com to register it as its own site.
```

The file is scaffolded on a site's first run and never rewritten afterwards. Site directories and the base URL stay owned by the site folder and cannot be overridden. Per-site configs apply to the global installation only — a project with its own config and `EXPLORBOT_*` env mode each resolve to a single config with nothing to extend.

## Surfacing

`explorbot sites` shows the config each site uses:

```
Registered sites (2):
  app.example.com    https://app.example.com    last run 2026-09-16 09:57
                     /home/you/.explorbot/sites/app.example.com/explorbot.config.js
  other.example.com  https://other.example.com  last run 2026-09-01 10:00
                     inherits global config
```

`explorbot config` and `explorbot api config` print it alongside the global one.

## Also in here

- **A config-cache bug.** `this.config` was populated before global mode finished, so a failed load left half-resolved settings behind. `ConfigCommand.summary` retries on failure, hit that cache, and reported success — which made the new validation unenforceable through `explorbot config`. Fixed in both parsers; regression test added.
- **CommonJS removed from config templates.** `init` wrote a CommonJS config for projects without `"type": "module"`, and `init --global` always wrote one. Explorbot is Bun-only and loads configs through `await import()`, so templates are now ESM throughout.
- **`deepMerge` deduplicated** into `src/utils/merge.ts` — it was copied in both config parsers.

## Testing

1434 unit tests pass. New coverage in `tests/unit/global-config.test.ts`: merge precedence, a site model spec resolving, both validation errors, scaffold-once, env mode not scaffolding, the cache regression, and the same for the api boat.

Verified end to end against a throwaway `HOME` for both boats: scaffolding, model override, and both error paths exiting 1.

Not covered by automated tests: none of the above needs a browser, so nothing here is gated behind the regression workflow.


DenysKuchma (changes requested):
superpowers/* must be remove?

DavertMik:
Later we will remove all superpower specs, so far no issues

PR #231, opened 2026-09-17
Run against a base URL that scopes paths and pins query params

DavertMik:
## Problem

A plan written against `/runs` cannot be run inside one project of a multi-tenant app. In global mode `enterGlobalMode` forces `playwright.url` to the site **origin**, so every relative path resolves at the root — `/runs`, never `/projects/acme/runs`. Putting the prefix in the base URL instead is not enough either: CodeceptJS joins `options.url + path` by concatenation, so the navigator then compares an expected `/runs` against an actual `/projects/acme/runs` and asks the AI to recover from a page it just loaded correctly.

Selecting a preview deployment by query parameter (`?assets_host=…`, `?preview=…`) had no mechanism at all: any relative navigation drops the query.

## What this adds

`--base-url <url>`, a common option on every browser command:

- **origin** — selects the site as before, so the site's config, knowledge and experience apply unchanged
- **path** — prefixed onto relative targets from plans and commands (`/runs` → `/teams/acme/runs`), idempotent, and never applied to absolute URLs so a login page outside the scoped area still works
- **query** — added to every document request to that origin, so a preview deployment stays selected for the whole run

```bash
explorbot test plans/runs.md '*' \
  --base-url 'https://app.example.com/teams/acme/?preview=pr-42'
```

## How it works

`ConfigParser` parses the flag once into `{origin, path, query}`: the origin becomes `playwright.url` (so CodeceptJS keeps joining against a clean origin), `applyBasePath()` owns the prefix rule, and `getBaseQuery()` exposes the params. Two seams call it — `resolveTargetPath()` for command targets, `Explorer.visit()` for the browser boundary — and `setCurrentPlan()` scopes a loaded plan's URLs in memory, so per-test start URLs, the `reset` tool and the AI's context all agree on the same URL. The plan file on disk is never rewritten.

The query is pinned with a Playwright route on the browser context that rewrites document requests only. The param reaches the server without entering the address bar, so state hashes, experience files and generated tests stay free of it.

## Verification

- `bun test tests/unit/` — 1448 pass, biome format + lint clean, `tsc` clean on changed files
- New unit test covers scoping, idempotency, absolute-URL passthrough and the origin-only `playwright.url`
- Checked against a live app: landing on a deep scoped URL delivered the preview assets, and navigation afterwards kept them

Not covered by unit tests: the route rewrite itself (needs a browser). Worth a regression label if you want it exercised end to end — I have not applied one.


src/config.ts
```diff
@@ -366,6 +367,7 @@ export class ConfigParser {
     if (this.config && !options?.config && !options?.path && this.runtimeTarget === target) {
       return this.config;
     }
+    this.runBase = parseRunBase(options?.baseUrl);
```
DenysKuchma, line 370: In config-free mode, EXPLORBOT_URL still wins when selecting the output root. With both values set, --base-url changes the browser origin but knowledge, experience, and output can come from the wrong site

bin/explorbot-cli.ts
```diff
@@ -82,6 +84,7 @@ function addCommonOptions(cmd: Command): Command {
     .option('--debug', 'Enable debug logging (same as --verbose)')
     .option('-c, --config <path>', 'Path to configuration file')
     .option('-p, --path <path>', 'Working directory path')
+    .option('--base-url <url>', 'Run against this URL: its path scopes relative paths from plans, its query params ride along with every page load')
```
DenysKuchma, line 87: This also exposes --base-url to rerun, but rerun executes saved I.amOnPage() calls directly through Mocha, bypassing the new path-scoping logic

DenysKuchma (changes requested):

